PR #6, opened 2017-04-13
Add forceDefaltIndex to budo config

rreusser:
Budo attempts to pick up any `index.html` in the cwd. I compiled `./index.idl` to `./index.html` (I mean should go in `build/`, but y'know, might not) and it broke the budo server. This PR adds [forceDefaultIndex](https://github.com/mattdesl/budo/blob/master/docs/api-usage.md#opts) so that it doesn't accidentally serve a static file.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -162,6 +162,7 @@ const start = () => {
   budo(path.resolve(__dirname + '/client/live.js'), {
     live: true,
     open: true,
+    forceDefaultIndex: true,
     css: '.idyll/styles.css',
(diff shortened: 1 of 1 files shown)


mathisonian:
Thanks! And thanks for teaching me about a new budo option

rreusser:
Np! Ran into it a bit when [using the master branch for src + web + _everything_](https://github.com/rreusser/smoothly-animating-points-with-regl), a strategy I'm starting to second-guess, but… 🤷‍♂️

PR #18, opened 2017-04-24
Update build process

mathisonian:
This PR starts work on the updated build process, outlined in this discussion: https://github.com/idyll-lang/idyll/issues/16. 

Still TODO:
- [x] Add code minification
- [x] Update https://github.com/idyll-lang/generator-idyll to use to use the new api 
- [ ] Update documentation
- [x] update tests

PR #19, opened 2017-04-24
Update styles and themes

mathisonian:
This PR does the following:

* Adds a new `github` theme (use it like `idyll index.idl --theme github`)
* Removes the `scroll` theme, it wasn't very good anyways and not mobile friendly. I can make a better one eventually, but for now better not to have it built in at all I think (you can get the same functionality using the `blog` theme anyways)
* Adds in a new `centered` layout. This layout differs from `blog` in that it doesn't put content off to the left and has fewer responsive breakpoints.
* Adds support for `none` theme and layout. Use by specifying `idyll index.idl --theme none --layout none`. In this case you are on your own for providing all of the CSS.
* ~Adds `browserify-css` so that components can include their own stylesheet without adding the styles inline. This helps with but doesn't completely solve https://github.com/idyll-lang/idyll-component/issues/1. One nice thing is that users will be able to use this transform in their custom components as well. It should also make it easier to override built-in component styles if they aren't done inline.~ I removed `browserify-css` in favor of `insert-css` since that is easier to work with and doesn't require an additional browserify transform.

screen shot 2017-04-24 at 4 36 04 pm

Example using `--theme github --layout centered`. I've updated the generator to make this the default configuration as well.

/cc @rreusser


mathisonian:
This requires https://github.com/idyll-lang/idyll/pull/18 to be merged to pass the tests

PR #28, opened 2017-05-01
Fix spellcheck arg parsing

rreusser:
Spellcheck could not previously be disabled. This PR fixes minimist parsing so that `--spellcheck` and `--no-spellcheck` both work, with a default of spellcheck: true.


mathisonian:
😅 thanks for catching this!

Issue #29, opened 2017-05-01
installing components from npm


mathisonian:
To me this feels like one of the last major features missing. 

Idyll components are react components are npm modules, so why can't I just `npm install` a third-party idyll component and use it in my project?

Here are a couple possible solutions:

### 1. Keywords, automatic scanning

Scenario:

1. A user publishes an idyll component to npm, they name the package: `idyll-component-a`. In the `package.json` they make sure to include the following:

```json
"keywords": ["idyll-component"],
...
"idyll": {
    "componentName": "componentA"
}
```

2. Another user has a local idyll project and runs `npm install --save idyll-component-a`. In their `index.idl` file they write:

```
[componentA /]
```

3. The idyll compiler picks up on this component because of the keyword tag, and everything works as expected.

### 2. Explicit listing

Scenario:

1. A user publishes an idyll component to npm, they name the package: `idyll-component-a`. It doesn't have any special configuration or keywords.

2. Another user has a local idyll project and runs `npm install --save idyll-component-a`. In their `package.json` file they list it explicitly: 

```
{
  "idyll": {
    "installedComponents": {
      "componentA": "idyll-component-a"
    }        
  }
}
```

In their `index.idl` file they write:

```
[componentA /]
```
and everything works as expected.

mathisonian:
One advantage of the second option is that you can install and use any react components (or similar), they don't have to be explicit idyll components. 

It is more work for the end user, but we could make it easier by providing some shortcut for it

```
idyll install componentA:idyll-component-a
```
and have that automatically updated `package.json`

rreusser:
Ooh. Yeah. I like part so of both. The two that I like are scanning for explicit triggers in the package.json and explicitly defining things in your package.json. The reason I like those is that it's less magic than relying on npm module names, which it seems a bit unfriendly to constrain with no escape hatch. If I had to take only one, I guess I think listing the component paths in your package.json is more flexible as you pointed out. The solution I'd use is both since they're not mutually exclusive. But I think a far better solution is to actually pick one. 😬 I do think listing them directly in your package.json is more flexible. Sure would be nice to automatically resolve though...

jbilcke:
Nice! thanks for asking me to join the conversation :)

I must admit I love the first option. It reduces the burden on the end user side, eases adoption, at the cost on some time investment in tooling and "magic". 

Maybe the workflow can ben even more simplified:

For the idyll module developer:
- should use a prefix (eg. idyll-component-*)  as a convention only
[...]

PR #32, opened 2017-05-03
Use components directly from npm

mathisonian:
See related discussion in #29 

This adds both automatic scanning, and a field in package.json to list explicit mappings if you prefer that. It doesn't add any `idyll install` or anything as I think its not really necessary anymore.

> I think this would be pretty hard to do without an explicit listing in package.json. The reason being that its hard to tell what module maps to a component vs. any other kind of library. That would lead to anything in node_modules being a candidate component which I think we should avoid.

I realized that this isn't entirely true, and so there wasn't such a strong argument for not scanning node_modules. This PR basically implements it as @bclinkinbeard described:

> User runs - `npm install secret-idyll-comp`
> In `.idl` file, user writes - [SecretIdyllComp /]

Idyll now has the following algorithm for module resolution:

1. Module mappings that are explicitly listed in package.json, e.g: 
```js
{
  "idyll": {
    "components": {
      "ComponentName": "./path/to/component", # requires a local file
      "OtherComponentName": "other-component-name" #requires from node_modules
    }
  }
}
```
2. Any component in the custom component folder
3. Any component in the default component folder
4. Any component found in `node_modules`

I'm pretty happy with where this added up, it feels nice to have no-config installable components. Thanks for the feedback @rreusser @jbilcke @bclinkinbeard

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,7 +16,7 @@ test_script:
   # Output useful info for debugging.
   - node --version
   - npm --version
-  - npm test
+  - npm run test-windows
 
 # Don't actually build.
 build: off
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "test-deps": "cd test/test-project && mkdir build && npm install",
     "test-local": "NODE_PATH=./test/test-project/node_modules/ mocha --timeout 20000",
-    "test-windows": "npm run test-deps && SET NODE_PATH=./test/test-project/node_modules/ && mocha --timeout 20000",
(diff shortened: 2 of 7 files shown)


mathisonian:
i'll squash this before its merged to get rid of this extra commits for the windows tests

mathisonian:
To note - for this to work Idyll needs access to the `node_modules` of the user's idyll project folder. If idyll is installed from `npm` like normal this is fine, but if you have idyll linked during development you'll need to set the `NODE_PATH` to make sure idyll scans that directory for possible components.  This is why the tests use `NODE_PATH` now https://github.com/idyll-lang/idyll/pull/32/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

PR #33, opened 2017-05-04
Fix up tests

bclinkinbeard:
This makes the tests run faster while ensuring the proper files are created with the proper contents.

`fs.readdirSync` returns an array, so `expect(Object.keys(outputFiles).indexOf(f)).to.be.above(-1);` could never fail, because running `Object.keys` on an array returns the indexes. `0`, `1`, etc.

I did have to update the JS and CSS files in the `expected-output` directory because they weren't matching the current output. I assumed they were just outdated.

I'm not sure why `yarn.lock` changed as the only thing I did was run `yarn install`.


mathisonian:
Does `postinstall` run for users installing this package as a dependency from `npm`? I'd like to avoid this if possible. 

I made a few changes to the `test` scripts in https://github.com/idyll-lang/idyll/pull/32/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 but I prefer what you've done here. I separated the tests for the windows CI because I couldn't figure out how to get the `NODE_PATH` working correctly there.

bclinkinbeard:
> Does postinstall run for users installing this package as a dependency from npm?

Yea, it would, which I agree isn't ideal. I guess we could just make a `quick-test` script or something that just runs mocha?

mathisonian:
Yeah, or maybe have `npm run init` set up all the test dependencies and everything and `npm test` just runs mocha? We'd have to update the CI scripts to use that but that isn't a big deal.

PR #37, opened 2017-05-05
Switch to Jest for testing

bclinkinbeard:
Apologies for the size of this change. The actual conversion from mocha to Jest is pretty simple, but in the course of trying to simplify the tests themselves I thought of a directory structure that should be more robust in the long term.

The structure is basically to have the `test` directory contain an arbitrary number of test project directories. Each test project directory (currently just `test/test-project`) would contain a `src` directory and an `expected-output` directory. That way there could be multiple test projects, each with their own independent configuration and set of expected results. Tests specific to that project could be housed in that directory as well, with more general tests residing in the root of `test`.

We'll probably need to find an alternative to `NODE_PATH` if we use this structure, but I think that's doable. Interested to hear your thoughts.


mathisonian:
I think this structure makes a lot of sense, but agree we should figure out an alternative to `NODE_PATH`. I can look into that.

One thing I also noticed - the expected output in the `.idyll` folder use fully qualified local paths, so those tests will fail unless we templatize them somehow.

bclinkinbeard:
I'm working on both of those right now. Pretty close, hoping to have an update in the next hour or so.

mathisonian:
Awesome! 

I also just noticed that the JS build step is using the `fullPaths` option, causing JS to get fully qualified paths too. This isn't necessary and can be removed https://github.com/idyll-lang/idyll/blob/master/src/index.js#L195

bclinkinbeard:
OK, the components are now required with relative paths, hooray! I still see a few absolute paths to the files in the `.idyll` directory in the `index.js` that gets generated, and I'm not sure where that happens.

bclinkinbeard:
I found the absolute paths and have a fix, just need to clean it up a bit first.

@mathisonian do you have any thoughts on the errors related to `idyll-component`?

mathisonian:
Great! The changes to the paths seems much cleaner, really nice to not have to worry about `NODE_PATH`. 

I'll confirm this, but from looking quickly I think the `idyll-component` error is coming from the fact that `npm run init` runs `npm install` inside of `test/test-project` but the `package.json` for that is now in `test/test-project/src/`.

mathisonian:
Yeah, that seems to be the issue, all tests are passing for me locally with that resolved (with the exception of the pending fixes to `index.js`). That dependency get included via `idyll-default-components`.

I set it up this way instead of just putting everything in `devDependencies` to make the test-project more closely match a real project. I guess this isn't totally necessary - we could just add `idyll-default-components` and `react-micro-bar-chart` (used for testing a component installed from npm) to `devDependencies`) and get rid of the additional install step.

src/index.js
```diff
@@ -129,16 +130,19 @@ const idyll = (inputPath, opts, cb) => {
...
         if (inputConfig.components[name]) {
-          outputComponents.push(`"${name}": require('${path.resolve(path.join(inputDirectory, inputConfig.components[name])).replace(/\\/g, '\\\\')}')`);
+          const compPath = path.parse(path.join(inputDirectory, inputConfig.components[name]));
+          outputComponents.push(`'${name}': require('${path.join(path.relative(TMP_PATH, compPath.dir), compPath.base)}')`);
         } else if (customComponents.indexOf(name + '.js') > -1) {
-          outputComponents.push(`"${name}": require('${path.join(CUSTOM_COMPONENTS_FOLDER, name).replace(/\\/g, '\\\\')}')`);
+          outputComponents.push(`'${name}': require('${path.relative(TMP_PATH, path.join(CUSTOM_COMPONENTS_FOLDER, name))}')`);
         } else if (components.indexOf(name + '.js') > -1) {
-          outputComponents.push(`"${name}": require('${path.join(DEFAULT_COMPONENTS_FOLDER, name).replace(/\\/g, '\\\\')}')`);
+          outputComponents.push(`'${name}': require('${path.relative(TMP_PATH, path.join(DEFAULT_COMPONENTS_FOLDER, name))}')`);
```
mathisonian: We need to keep these `.replace(/\\/g, '\\\\')` to make the tests pass on windows

bclinkinbeard: Ah OK, will do that now.

bclinkinbeard:
Hmm, that seems to have broken Travis and didn't fix Windows. Did I do it wrong?

mathisonian:
No that looks right to me - the travis error is a timeout (it passed on node 6 but not 7) so I'm wondering if we're just close to jests timeout? It's still passing for me locally, takes about 14 seconds

mathisonian:
From the jest docs:


> If your test is long running, you may want to consider to increase the timeout specified in jasmine.DEFAULT_TIMEOUT_INTERVAL.
> `jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000; // 10 second timeout`

bclinkinbeard:
Welp, I'm not sure what Windows needs to be happy. Based on a quick scan of the output it looks like maybe comparing the file contents isn't going to work on Windows? No clue tbh.

I'm wrapping up for the evening (I'm on Eastern time) but feel free to tweak things as needed if you have ideas.

mathisonian:
I don't really know whats going on with comparing output on windows, but with a couple tweaks everything else is passing so I've gone ahead and merged.

PR #39, opened 2017-05-09
Refactor - pass artifacts to callback

bclinkinbeard:
This ended up being a lot bigger than I expected. Sorry about that (again), it just ended up touching more places than I thought to unwind the filesystem assumptions.

That said, I'm pretty happy with how things worked out. It's far from perfect, and I have additional improvements in mind that I will submit as separate, smaller PRs, but I think these changes prepare the codebase for easier iteration going forward. The major theme of this PR is separating the AST processing from the writing of artifacts to disk. You'll notice the files are now written in their own small functions, and the code that constructs the file contents is unaware of how those contents are used/persisted.

The processed results (artifacts?) are now also returned to the callback if you pass one to `idyll()`. I *think*, with a few more iterations, we could make the filesystem access completely optional. That would of course enable the entire process to run in the browser and allow us to avoid roundtrips to the server when working in a browser based editor.

Two disclaimers: This PR doesn't yet address the start/watch workflow. I will work on that while these larger changes are reviewed, so don't merge this before those updates land. Also, I'm sure we'll want to work on performance once these changes are merged.


src/index.js
```diff
@@ -79,168 +75,166 @@ const idyll = (inputPath, opts, cb) => {
...
-          COMPONENT_FILE: path.join(path.relative(visitorsDir, TMP_PATH), path.parse(COMPONENT_FILE).base),
-          DATA_FILE: path.join(path.relative(visitorsDir, TMP_PATH), path.parse(DATA_FILE).base),
-          IDYLL_PATH
-        } ],
         [ brfs ]
       ]
     });
+    b.require(AST_FILE, {expose: '__IDYLL_AST__'});
+    b.require(COMPONENT_FILE, {expose: '__IDYLL_COMPONENTS__'});
+    b.require(DATA_FILE, {expose: '__IDYLL_DATA__'});
```
bclinkinbeard: Instead of relying on env vars, Browserify allows us to basically [alias these dependencies](https://github.com/substack/node-browserify#brequirefile-opts). This avoids some of the path craziness we had, removes the dependency on `envify`, and of course removes one more thing tying us to Node.

rreusser: Ooh, nice trick. There's also a fun [budo plugin trick](https://github.com/mattdesl/budo/issues/201#issuecomment-285160155) I've been using for budo's browserify instance.

bclinkinbeard: Nice! I'm about to tackle the budo stuff so this could be super handy. I was pretty stoked when I figured out the expose stuff because I'd been struggling on that piece. :)

mathisonian: I guess we could eventually require a stream directly here and avoid creating the file altogether?

mathisonian: I'm not super eager to do this though as having the  `AST` and `COMPONENT` file has been really useful for debugging

src/index.js
```diff
@@ -79,170 +75,174 @@ const idyll = (inputPath, opts, cb) => {
...
+            } else if (components.indexOf(name + '.js') > -1) {
+              acc[name] = path.relative(TMP_PATH, path.join(DEFAULT_COMPONENTS_FOLDER, name)).replace(/\\/g, '/');
             } else {
-              data = require(path.join(DATA_FOLDER, source));
+              try {
+                // npm modules are required via relative paths to support working with a locally linked idyll
+                const compPath = path.parse(resolve.sync(name, {basedir: inputDirectory}));
+                acc[name] = path.join(path.relative(TMP_PATH, compPath.dir), compPath.base).replace(/\\/g, '/');
+              } catch (err) {
+                if (node[0].toLowerCase() !== node[0]) throw new Error(`Component named ${node[0]} could not be found.`)
```
mathisonian: This shouldn't throw an error - components that don't match anything shouldn't be included in the `COMPONENTS_FILE`,  but should still be allowed

bclinkinbeard: Won't that mean components that can't be found will fail silently?

mathisonian: The `require` should fail silently, just indicating that a component with that name isn't installed. The client side code knows how to handle components that aren't defined (see resolution algorithm here https://idyll-lang.github.io/components-overview)

bclinkinbeard:
Do we need anything else here or can I push the button? 😆

mathisonian:
Should be good to go 👍

PR #41, opened 2017-05-10
Check in test project deps

bclinkinbeard:
Test project build artifacts *will* be ignored now, since they should be generated fresh every time the tests run. Artifacts will also be cleared at the beginning of each test run.


mathisonian:
Cool, this means we can remove the custom windows test script as well

bclinkinbeard:
Oh cool, I can push up a commit for that. I am again mystified that Travis builds are failing though.

bclinkinbeard:
What just happened lol

mathisonian:
I updated the CI scripts to use yarn, I think that + `node_modules` should fix the build

PR #47, opened 2017-05-10
fix if there are no meta nodes

### LINKED ISSUE #44: error is thrown if there is no meta component
The `getMeta` function should be updated to handle the case where no meta component is defined

mathisonian:
fixes #44


bclinkinbeard:
LGTM

Issue #40, opened 2017-05-09
JS and CLI API discussion


bclinkinbeard:
The current API looks like this in JS: `idyll(inputPath, opts, cb)` where `inputPath` is the absolute path to your `.idl` file, `opts` is an options object, and `cb` is a callback function that will receive the build artifacts once #39 lands.

The CLI is of course very similar, since it calls the JS API, looking something like this: `idyll index.idl --layout centered --css styles.css --build`, where `inputPath` is the first argument and everything else is translated to a property on the options object.

Of note:
- the CLI accepts relative paths 
- there is no analog to the callback when using the CLI

Resolution of `inputPath` should definitely move into the JS part so `inputPath` can be a relative path regardless of how Idyll is being used.

**Questions**
- Should the CLI return a stream or something to support piping into something else? Seems low priority.
- How should we support passing in Idyll source, rather than forcing the reading of a `.idl` file from disk?

One approach that seems fairly simple would be to support the omission of `inputPath` entirely, and in that scenario look for a `inputSrc` (or similar) property on `opts`. This is similar to how Browserify allows you to [specify `opts.entries`](https://github.com/substack/node-browserify#browserifyfiles--opts) instead of providing `files` as the first argument to the `browserify()` function.

Thoughts @mathisonian @rreusser et al?

mathisonian:
I think updating it to something like 

```
var inputOptions = {
   inputString: ''
};

idyll(inputOptions, cb);
```

makes sense. I'm also in favor of moving all the file writing logic into the `idyll` bin, and out of the idyll javascript command itself. 

It doesn't seem necessary for the CLI to return a stream at this point. I think we should encourage people to use the API directly in that case.

bclinkinbeard:
@mathisonian you mean make the `idyll()` function ALWAYS take an options object as the first parameter, and an optional callback as the second? And the options object would need to have either `inputPath` or `inputString` set?

Would that mean the CLI would look like `idyll --inputPath index.idl`?

mathisonian:
Yeah, in that case it would always have options as the first argument. 

I'd leave the CLI as is - `idyll index.idl` - since that is a bit more in line with how other CLI tools work and seems more intuitive. I don't see it as an issue that the command line tool is slightly different than the javascript api.

bclinkinbeard:
Cool, works for me

PR #50, opened 2017-05-11
Update CLI and API

bclinkinbeard:
Implements the changes discussed in #40.

Also includes a good amount of path code cleanup. I think it's a good idea to centralize and quarantine the code that deals with paths as much as possible. The CLI doesn't do anything with paths now, its sole responsibility is providing and documenting the CLI API. It doesn't do anything besides passing the options to the JS function.

This also brings all of the various ways to specify options into harmony by way of matching names. Options specified on the command line and/or passed in the options object to the function override options specified in package.json override default options. See the `alias` method in the bin file for a list of option names that is easy to read and review.


bclinkinbeard:
Added a commit to remove the `package.json` config from the override cascade since `components` means something different in that context.

bclinkinbeard:
Two questions have to be answered before we can land this. 

1. Default to file watching with a `--build` option or default to a single build with a `--watch` option?

2. Minify by default with a `--debug` option or don't minify by default with a `--prod` option?

I vote for `--watch` and `--prod` as they are common across a lot of tools.

mathisonian:
Thanks @bclinkinbeard! 

I'm happy to switch over to opting in to `watch`, for the JS api this seems clearly the right choice. For the CLI its less obvious, but I think its nice to have consistency, and is in line with how some other tools behave. 

For minifying - I'd like to keep that opt-out for the build step. One thing that Idyll does now that I'd like to keep is to default to best-practices unless you explicitly opt out of them. I'd also vote to keep the option named as `minify` for clarity unless you forsee other things getting rolled into that in the near future.

src/index.js
```diff
@@ -25,69 +31,81 @@ require('babel-core/register')({
...
+const idyll = (options = {}, cb) => {
+  const inputDir = options.inputFile
+    ? isAbsolute(options.inputFile)
+      ? dirname(options.inputFile)
+      : dirname(join(process.cwd(), options.inputFile))
+    : process.cwd();
+  const getPath = getPathFactory(inputDir);
+  const inputPkg = require(getPath('package.json'));
+  const inputCfg = (inputPkg.idyll || {components: {}});
+  for (let key in inputCfg.components) {
```
mathisonian: Small style thing - can we keep things like `inputCfg` spelled out like `inputConfig`, this makes the code more friendly and readable.

bin/idyll.js
```diff
@@ -1,53 +1,67 @@
...
-    options.theme = path.resolve(options.theme);
-  }
-}
+// delete stuff we don't need
+delete argv._;
+delete argv['$0'];
+delete argv.k;
+delete argv.spellcheck;
 
-idyll(path.resolve(argv._[0]), options);
```
mathisonian: I added this `path.resolve` in to resolve some weird bug when using idyll linked locally. Will test this to make sure everything still works.

bin/idyll.js
```diff
@@ -1,53 +1,67 @@
...
+  .default('output', 'build')
+  .boolean('spellcheck')
+  .default('spellcheck', true)
+  .describe('spellcheck', 'Check spelling of Idyll source input')
+  .describe('template', 'Path to HTML template')
+  .default('template', '_index.html')
+  .describe('theme', 'Name of (or path to) the theme to use')
+  .default('theme', 'idyll')
+  .alias('h', 'help')
+  .argv;
```
mathisonian: 👍 so much cleaner

mathisonian:
We'll need to update the project generator too to match this

mathisonian:
I'm getting some weird behavior locally -

This is running the script with idyll linked locally:

```
    "start": "idyll index.idl --css styles.css --layout centered --theme github --spellcheck"
```

* It seems like the compiler is being triggered a bunch of times on startup (spellcheck keeps re-running)
* The styles are 404'ing 

idyll-api-updates

src/index.js
bclinkinbeard: Yep, will fix that

bclinkinbeard:
Hmm, I admittedly use `npm run build` most of the time when testing so it's possible I overlooked something. I will check into that now.

Regarding the CLI, you want to minify by default and turn it off with `--minify false`?

bclinkinbeard:
Or maybe `--no-minify` is better?

mathisonian:
👍 `--no-minify`

bclinkinbeard:
All issues addressed, except the `npm start` thing as that is coming from the `chokidar` stuff in `master`. I will address that in another PR.

PR #52, opened 2017-05-12
fix reloading issue introduced with chokidar and CSS bug

mathisonian:
This updates the watcher to wait for the `ready` event, so that an `add` event isn't fired initially for every file in the watched directories


mathisonian:
There was also a bug in the live CSS path, this PR fixes that as well so budo receives `build/style.css` instead of `styles.css` during live development.

Issue #31, opened 2017-05-03
propagate errors from compiler into the browser


mathisonian:
a great feature of budo is that it shows errors directly in the browser so you know when something isn't compiling and it doesn't get lost in the terminal. idyll should do this too.

mathisonian:
I've added initial support for this in https://github.com/idyll-lang/idyll/tree/error-messages but have a lot of clean up to do!

joshuahhh:
Why is this issue closed? I don't think it was implemented – I still get errors in the console, rather than in the browser.

FWIW, this is my #1 workflow problem with Idyll. I am going back and forth between my editor and the browser, and then suddenly the browser stops updating, and I think it's because something is broken, but it turns out it's because there's some syntax error hiding in the console. It's quite frustrating!

(Thanks!)

mathisonian:
Thanks for pointing this out @joshuahhh - I can confirm this is still an issue. 

I'm working on a patch.

joshuahhh:
That's great – thanks so much!

Not sure what you were planning, but my favorite-to-least-favorite ordering for interfaces is:
1. an impossible-to-miss but otherwise unobtrusive overlay on the document,
2. replacing the document with an error message,
3. (way worse) a message in the console.

Maybe unrelated: I've found that some errors (syntax errors in custom-component Javascript, maybe?) straight-up exit Idyll. Should I make a new issue for that?

mathisonian:
Yes, could you make a separate issue for those?

mathisonian:
Closed by #458

PR #54, opened 2017-05-12
Pipeline overhaul bonanza

bclinkinbeard:
This should be my last massive PR, hopefully ever.

This completely breaks the build pipeline out into its own set of discrete steps, and moves path creation/config into its own module as well. This granularity will allow us to evolve (and test!) various pieces of the system in isolation. 

As an example, creating the JS bundle is now the sole responsibility of `src/pipeline/bundle-js.js`. It will be easy to iterate on the bundling process going forward to add better caching, split bundles, etc. This single bundling config is also used in both watch and normal modes now. Bundling takes about 2-3 seconds right now, but I'm confident we can get that down closer to 1 second.

This setup will also make it easy to add steps to the pipeline if we want to, so things like tree shaking, asset optimization, etc. could be added without disturbing other parts of the system.

The one thing I just realized I did not put back in here is CSS updates and injection. That will be easy enough, but I wanted to at least get this submitted before the weekend. You can let me know if you'd prefer I add that to this PR or do it separately after this is merged.


bclinkinbeard:
Woohoo, just added CSS injection. That was easy.

mathisonian:
This is awesome! I'm about ready to merge, but just have a few things:

  * I've never used browsersync before - could you post a little about your reasoning moving to that tool?
    * I added `{notify: false}` to the browserSync startup options, this just removes that "connected to browsersync" message. IMO idyll users shouldn't really need to know about or care about the underlying tech serving the page.
    * Right now if the compiler throws an error everything crashes. I think this should be as simple as adding an exception handler to the `pipeline.preBundle` promise. It would be awesome if we can pipe the message into the browser re https://github.com/idyll-lang/idyll/issues/31. It seems like that can be down with `browserSync.notify()`
    * Do you know if there's an easy way to customize the browsersync promp in the command line?
  * This subtly changes the static asset handling (before the base directory during live dev was the root of the project folder rather than the build folder). This seems fine to me, but something to note. I think we should add some logic to make it easier to deal with a folder of images for example, but lets leave that as a separate thing. 
  * Access to intermediate values in the pipeline - I'm thinking of updating the compiler to return for example spellcheck and other metadata in a more structured format. it seems like it would be slightly difficult to access that since its in the middle of the pipeline if we e.g. wanted to pass it into browsersync.

bclinkinbeard:
Cool, lots of good points. I went with BrowserSync just because it seems like a more general purpose tool for this sort of thing. budo seems aimed at being a simple way to set up a server with watchify integrated, but I think it will serve us better long term to be able to watch various paths and react in certain ways, etc. BrowserSync also lets you connect multiple devices and mirror things like scrolling and clicks across them, so developing a piece for desktop and mobile should be very easy.

> Right now if the compiler throws an error everything crashes
Good catch. I meant to circle back and handle errors and forgot.

> It would be awesome if we can pipe the message into the browser
Agreed, will work on this.

> easy way to customize the browsersync prompt in the command line?
Not sure, will check

> before the base directory during live dev was the root of the project folder rather than the build folder
I didn't realize that but maybe that would be better? That would allow us to serve images and things without copying or aliasing them. Want me to revert to the previous base directory behavior?

> Access to intermediate values in the pipeline
Also a good idea. Maybe we should restructure what gets passed to the callback. Maybe just use a traditional Node callback where `err` is the first arg? Open to suggestions. I will also ensure the callback works in watch mode, I think that got broken.

mathisonian:
Sounds good 👍 

> Want me to revert to the previous base directory behavior?

Let's keep it how it is to get the PR landed and address that separately

bclinkinbeard:
Addressed all feedback in latest commit. Things are slightly messy but well encapsulated in the pipeline.

The only change it seems you can make in BrowserSync's logging is to set the [`logLevel`](https://www.browsersync.io/docs/options#option-logLevel) and [`logPrefix`](https://www.browsersync.io/docs/options#option-logPrefix), which I did. It will now prefix messages with "Idyll" instead of "BS" and will only log warning messages.

Items to address separately after this PR lands: which directory should be served as the server root in watch mode and how do we want to alter the callback signature and what gets passed to it.

PR #56, opened 2017-05-18
POC browserify-incremental integration

bclinkinbeard:
Work in progress/proof of concept. Do not merge.

I just wanted to get it on record what this looks like. Things to note:

- `.pipe()` is required, which is undocumented and annoying
- I suck at streams so I just piped it to the output file for now
- does not honor `minify` setting
- requires `fullPaths: true` which would necessitate changes in how we test


bclinkinbeard:
Closing in favor of #57

PR #57, opened 2017-05-18
Switch to an instance and event based API

bclinkinbeard:
This PR switches the API to what we discussed on Gitter.

```javascript
const idl = idyll(config);
idl.on('update', doStuffWithOuput);
idl.build(); // nothing happens until build() is called
```

This also ditches `watchify` in favor of `browserify-incremental`. Watchify made the pipeline more complicated, required a long lived process for fast builds and, worst of all, seems to have some caching issues. Despite a lot of debugging to ensure all the dependent files were being updated as expected, watchify would sometimes produce stale builds. The only downside to `browserify-incremental` that I know of is it requires `fullPaths: true`, which means we can't compare build output across machines. Tests have been updated/disabled accordingly.

The event capabilities are currently implemented by creating a `IdyllInstance` class that extends `EventEmitter`. We don't have to do it that way but it was the quickest and most straightforward. BrowserSync is only used if `options.inputFile` is set (and `options.watch` is `true`), since that indicates a local workflow with a need to refresh browsers. I'd like to clean up both of these aspects of the code but for now they work. The pipeline can also be cleaned up and simplified because we only use the `build` method now, which is a huge win.

The only bug I haven't been able to fix in this code is that when you are running it with BrowserSync enabled, the event listener only fires once. It is somehow removed after the initial build. This is weird, and I want to figure it out, but it's unlikely you'd be using BrowserSync AND a listener, so I doubt people would actually run into this.

<details>
<summary>Check out the sweet server code this enables</summary>

```javascript
const idl = idyll({
  inputString: fs.readFileSync(tmpFile, 'utf8'),
  output: 'omfg',
  minify: false,
  debug: true,
  compilerOptions: {spellcheck: false}
});

app.post('/', (req, res) => {
  idl.once('update', (stuff) => {
    res.json(stuff);
  }).build(req.body.src);
})

app.listen(3001, () => {
  console.log('Go go go!');
})
```
</details>


bclinkinbeard:
Latest commit really boils the pipeline down to as simple as it can be. Everything prior to the JS bundling is actually synchronous, so now the code treats it as such, making it much more readable.

Still left:
* handle compilation errors again
* figure out what to do with `updateCSS`. Maybe `build` is fast enough that we don't need it?
* determine which files are no longer needed and delete them

bclinkinbeard:
Items mentioned above have all been addressed and this can be merged once approved

PR #61, opened 2017-05-26
Server files from input *and* output directories

rreusser:
**Observed behavior**: browser-sync fails to serve files from `data/*`, `images/*`, `fonts/*` etc unless the project has already been built. 

**Expected behavior**: the dev server serves files from your source directory

**Solution**: This PR adds the input directory to the browser-sync static directories

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -77,7 +77,7 @@ const idyll = (options = {}, cb) => {
               logLevel: 'warn',
               logPrefix: 'Idyll',
               notify: false,
-              server: paths.OUTPUT_DIR,
+              server: [paths.INPUT_DIR, paths.OUTPUT_DIR],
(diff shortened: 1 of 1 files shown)


rreusser:
A small note: for some reason the output directory has to come first otherwise themes and layout seems to break, which is too bad because it seems like it'd be best if the dev server weren't using the filesystem at all. This directory order seems to work and make local assets available though, though I suspect it's going to b pretty confusing to fight between build and development assets.

mathisonian:
@rreusser - the reason for that is that order mattering is the user defined styles get combined with the theme and layout styles into one stylesheet and put into the build directory. i agree it'd be better to do everything with streams, we're moving in that direction with all of @bclinkinbeard's work

Issue #11, opened 2017-04-13
glslify support [and other custom transforms]


mathisonian:
It would be nice to make the transforms passed to browserify configurable, or at least let people add more in addition to the ones that idyll includes by default.

PR #63, opened 2017-05-26
Support user-defined transforms

### LINKED ISSUE #58: NODE_ENV shouldn't be set automatically during development
`NODE_ENV=production` should be set automatically during a build, but this shouldn't be set if `watch` is true: the production flag will cause valuable debugging information to become inaccessible.

mathisonian:
This PR allows users to specify custom browserify transforms to run in addition to the default ones provided by Idyll, fixing #11. 

It exposes three properties in the Idyll config (currently accessed through `package.json`) that manipulate the browserify bundle transform chain: `pretransform`, `transform`, and `posttransform`. The reason for including all three is that the order can be finicky sometimes. 

This enables, for example, use of `glslify` in custom components by adding this to the `package.json` file:

```json
  "idyll": {
    "transform": [
      "glslify"
    ]
  }
```

It also fixes #58.

One thing to note - this doesn't allow users to wholesale replace the transforms that idyll currently has set up as that would require a bit more complicated specification, and I think I'd rather wait and see if that is requested to add it.

cc @rreusser @bclinkinbeard


bclinkinbeard:
This seems a bit overly complex to me. Couldn't we just stick any custom transforms in the middle of the two defaults? I don't think anything should need to run before `babelify` or after `brfs` should it? Also, I think we'd want to make sure any options can be provided via any of the 3 supported mechanisms: `package.json`, CLI, or API.

I'm also curious about what (if anything) this assumes about how custom components are used. Is this using the assumption that a custom component will be in the local `components` folder and not inside `node_modules`?

rreusser:
I agree, though I don't generally have a problem with obscure extra flags like `pretransform` that have no effect unless you *really* need them—as long as the implementation is trivial.

mathisonian:
@bclinkinbeard these transforms would only be applied to components outside of `node_modules`, presumably things inside of `node_modules` will encapsulate their own transforms

mathisonian:
I've updated the PR based on the feedback so `transform` can now also be passed via command line or via the API in addition to the `package.json`.

I simplified it to remove pre and post transforms, but those should be straightforward to add back if there is demand for it.

PR #64, opened 2017-05-26
depend on the default component repo

mathisonian:
this should be merged alongside https://github.com/idyll-lang/generator-idyll/pull/5


bclinkinbeard:
I'm confused by this, it doesn't help us until/unless we add a `prepublish` script, right?

mathisonian:
It doesn't solve the issue of using them directly from `node_modules`, just moves the dependency from the generated projects to here so I can update the generator in light of the upcoming changes

Issue #34, opened 2017-05-04
Markdown support is incomplete and inconsistent


jheer:
The use of Markdown syntax is a wonderful way to leverage people's existing familiarity. However, there are some gaps in coverage; examples included below. Some should be easy to address, others perhaps less so. At minimum, it seems important that the documentation eventually note any discrepancies, which seem likely to trip up new users.

### Missing Constructs

- Italic/bold styling using underscore (`_italic_`, `__bold__`).
- Unordered lists initiated by hyphen (`- List item`)
- Ordered lists (`1. List item`)
- Syntax highlighted code blocks (```` ```json ````)

### Inconsistent Constructs

- Using an asterisk as the first character of a line in Idyll results in an unordered list. Other Markdown flavors will produce italicized text if a line of text consists solely of `*italic*`, and bold text if a line consists solely of `**bold**`.
- Hyperlinks `[text](url)` collide with Idyll's component syntax. Trying to include a Markdown hyperlink can result in an Idyll compilation error, which could be jarring for some. Perhaps the use of parentheses and lack of closing bracket `/]` is syntactically unique here? If so, standard links might be supported without modification to component definitions.

mathisonian:
Thanks @jheer - 

I have issues for most of these in the [compiler repo](https://github.com/idyll-lang/idyll-compiler), but I think its good to keep this issue open until we reach a reasonable parity or have the differences well documented. 

The biggest thing right now is restructuring how the asterisk characters are parsed to make it more consistent with other markdown implementations, as you note. This is being tracked in https://github.com/idyll-lang/idyll-compiler/issues/3, I think once that is done `Italic/bold styling using underscore` and `ordered lists`, and `Unordered lists initiated by hyphen` should follow fairly easily. 

I've just opened a PR addressing markdown style hyperlinks, and will open a separate issue for `Syntax highlighted code blocks`.

mathisonian:
https://github.com/idyll-lang/idyll-compiler/pull/14 addresses most of these issues. The outstanding ones are:

* Syntax highlighting (tracked in https://github.com/idyll-lang/idyll/issues/38)
and
* Having clear documentation of differences from markdown (tracked in https://github.com/idyll-lang/idyll-lang.github.io/issues/1)

rreusser:
Added bold *and* italic via idyll-lang/idyll-compiler#28

PR #65, opened 2017-05-26
add support for syntax highlighting

### LINKED ISSUE #38: language specific syntax highlighting for code blocks
https://github.com/idyll-lang/idyll-compiler/pull/14 adds support to capture a language associated with a code block, but this information isn't used right now. 

Idyll should automatically do syntax highlighting if you specify a language.

mathisonian:
fixes #38 and related to #34.

This adds syntax highlighting via https://github.com/conorhastings/react-syntax-highlighter. It parses the input to determine which language highlighters are needed, and builds a bundle such that only the necessary ones are brought in. 

This works by instantiating a new [code-highlight](https://github.com/idyll-lang/idyll-default-components/blob/master/components/code-highlight.js) component.


mathisonian:
I think we just need to merge #64 to get the tests to pass

src/pipeline/parse.js
```diff
@@ -108,6 +108,49 @@ exports.getDataJS = (ast, DATA_DIR) => {
   return `module.exports = ${JSON.stringify(data)}`;
 }
 
+
+exports.getHighlightJS = (ast) => {
+
+  const languageMap = {
+    js: 'javascript'
```
bclinkinbeard: Does this mean you could have <pre>```js</pre> in your Idyll source? That would be awesome, and it's what I always wish was standard, but we should probably support <pre>```javascript</pre> too.

bclinkinbeard:
LGTM other than the small comment above.

src/pipeline/parse.js
mathisonian: yeah exactly. if the language isn't in `languageMap` it will still get passed in as-is, so 

````
```javascript 
````

would still work

Issue #55, opened 2017-05-15
Eliminate need for project files


rreusser:
It would be nice if (like markdown), you could just open up a single file and start working. The minimal project for me looks like:

```
project-dir
├─┬ components
│ └── default_components
├── package.json
└── index.idl
```

(It's a little unhappy that `_index.html` doesn't exist, but it still seems to render.) Seems like it would maybe be nice (and not fundamentally difficult?) to eliminate the need for `components` and `package.json` so that you really could just open up a file and start, then add pieces as needed. Presumably it would just override built-in defaults with more pieces as specified. This is probably more a matter of configuration than anything else.

The benefit to me is that it's easier to start from scratch and feels more like a 'language' + 'document' when it's just a file and pulls in external resources as needed, rather than the fundamental unit being a carefully organized collection of files. (= framework?)

Especially since #50, looks like shouldn't be too bad. Thoughts/opinions?

mathisonian:
I'm in favor of this. The work breaking apart the build process should help. 

I think the biggest change needed would be to allow for `idyll` to fallback to its own copy of default components (and `_index.html`?) if none are specified.

rreusser:
The only weird part seems like maybe managing relative vs absolute paths in local vs default components, but I bet that can be solved.

PR #66, opened 2017-05-27
Batteries included

bclinkinbeard:
This enables use from vanilla projects not created with the generator by taking advantage of the newly built-in components and removing a couple of assumptions about project structure. Using this code you can do this in a fresh, blank project:

`require('idyll')({ inputFile: 'input.idl' }).build()`

This also fixes a "bug" introduced in #65 where it was assumed that source projects would have a dependency on `react-syntax-highlighter`.


rreusser:
Apologies if it's outside the concern of this PR, but is it possible/reasonable for `inputFile` to be a string or a stream or do you have to back up and assemble the pieces yourself if you want to go to that level?

bclinkinbeard:
@rreusser you can provide a string as `inputString` but we don't currently support streaming.

bclinkinbeard:
Latest commit updates the CLI to match the new (lack of) default options. This means that, if you had `idyll` installed globally, you could simply create an `index.idl` file and run `idyll index.idl --watch` and be off to the races. This is the final piece in making #55 a reality!

rreusser:
Is it possible to modify [this section](https://github.com/idyll-lang/idyll/blob/d2ee8f79e8725954968e5875f3e89be402ca951d/src/index.js#L47-L55) to proceed gracefully if there's no `package.json`? In fact it seems like it'd be better to pass the `package.json` content through options instead of requiring a local file from within this section.

It still fails for me without `package.json`, so this seems like perhaps the last modification necessary to make it truly batteries included.

bclinkinbeard:
@rreusser Good catch. I didn't understand the part about passing it around but I added a check to make `package.json` optional.

src/index.js
```diff
@@ -44,7 +45,7 @@ const idyll = (options = {}, cb) => {
 
   const paths = pathBuilder(opts);
 
-  const inputPackage = require(paths.PACKAGE_FILE);
+  const inputPackage = fs.existsSync(paths.PACKAGE_FILE) ? require(paths.PACKAGE_FILE) : {};
```
rreusser: Thanks! 🎉

rreusser:
@bclinkinbeard I was wondering what it would mean to use idyll client-side. Part of that is collecting all the fs operations into one location so the other parts don't depend on a local fs, but this file probably doesn't apply at all if you're on the client.

bclinkinbeard:
Agreed that quarantining all `fs` ops would be a good idea, it made me cringe to add it there. I'm OK with kicking that can down the road though since actually eliminating the `fs` dep is a huge ordeal far off in the future. ¯\_(ツ)_/¯

rreusser:
Yup, no worries here. 😄 It has a hard dependency on fs access anyway, so nothing lost IMO. Wondering if it would someday be possible to write `require('idyll')({ inputString: '...' }).build()` and get the output without fs needing to assume anything about fs access, but yeah, not a goal currently, I think.

mathisonian:
I'm good to merge this as soon as I can figure out how to get the tests to pass on windows

mathisonian:
I'll update the generator to explicitly set the `defaultComponents` path, and we should probably pull the `_index.html` from Idyll now for consistency.

PR #68, opened 2017-05-27
[WIP] Build to static output

rreusser:
This PR starts working in the direction of building to static output. For one, it makes the simple change of passing `ast` to `InteractiveDocument` instead of requiring it from within. That makes it trivial to pass the AST to the component when building to static or to use the fancy `require('__IDYLL_AST__')` trick from *outside* the component when building the client bundle.


mathisonian:
Closing this since #75 was merged. thanks for the guidance on this @rreusser

PR #75, opened 2017-06-03
SSR POC

bclinkinbeard:
Quick and dirty experimentation following the lead of @rreusser. Still plenty to do as it bombs when trying to create the bundle, but it does pre-render the HTML file. (Not positive it's correct since I can't test it but it looks to be right.)

One of the most obvious issues is pathing, which is starting to feel like a pattern. :) The way things are currently set up, with the bundling reading files from the `.idyll` directory, the paths needed for that process are different than the paths needed for SSR. If we can unify that, preferably by avoiding those temp files altogether, I think we'll be pretty close to having this working.


src/client/build.js
```diff
@@ -2,6 +2,17 @@ const React = require('react');
...
+const mountNode = document.getElementById('idyll-mount');
 
-ReactDOM.render(<InteractiveDocument />, mountNode);
+const ast = require('__IDYLL_AST__');
+const componentClasses = require('__IDYLL_COMPONENTS__');
+const datasets = require('__IDYLL_DATA__');
+
+ReactDOM.render(
+  <InteractiveDocument
+    ast={ast}
```
rreusser: 👏

src/client/component.js
```diff
@@ -3,8 +3,7 @@ const walkVars = require('./visitors/vars');
 const walkNode = require('./visitors/node');
 const utils = require('./utils');
 
-require('__IDYLL_SYNTAX_HIGHLIGHT__');
-let results = require('__IDYLL_AST__');
+// require('__IDYLL_SYNTAX_HIGHLIGHT__');
```
rreusser: I didn't investigate what this one does but maybe needs an alternate implementation?

mathisonian: This executes some code that makes only certain languages available to the syntax highlighter to avoid bloating the bundle. Maybe we could just require the entire lib during SSR since we don't need to worry about filesize there.

rreusser: Yeah, that would work, I think. *Possible* via a conditional require and then `envify` to strip it out so that it's removed for the client bundle. Otherwise might have to do the same trick and pass it down from the top.

bclinkinbeard:
I rebased this on #76 and now the `test-project` test seems to hang.

mathisonian:
Hmm it looks like an error coming from (loaded in the custom component in `test-project`)

```
Unhandled rejection ReferenceError: window is not defined
    at /Users/conlenm/projects/idyll-lang/idyll/node_modules/scrollwatch/dist/ScrollWatch-1.2.0.min.js:2:185
    at t (/Users/conlenm/projects/idyll-lang/idyll/node_modules/scrollwatch/dist/ScrollWatch-1.2.0.min.js:2:107)
```

mathisonian:
I'm working on a fix for 👆

mathisonian:
that should be fixed in https://github.com/idyll-lang/idyll-component/commit/f5871833cb5acd0399e8cbd56b3d078fd33a9374, and with the latest commits this works on local projects for me, but that test is still hanging for some reason.

mathisonian:
I fixed a couple minor issues causing the tests to hang, everything should be passing now.

bclinkinbeard:
Latest commits cleans things up a bit by moving the string conversions (wrapping with `module.exports = ` etc) into the bundling step. That allows the `getHTML` step to work with the raw objects to enable SSR.

The tests are broken because of the change in output, but those will be easy fixes if we like this approach to the code.

`react-syntax-highlighter` is also still causing some issues. No matter the language specified there is a RTE saying the language can't be found.

mathisonian:
That seems like a great approach, I was going to suggest something similar 👍  I've updated the tests, and am looking into the runtime errors with syntax highlighting

mathisonian:
I think we should probably only enable this if `watch` is `false`?

PR #76, opened 2017-06-05
Less reliance on the filesystem

mathisonian:
This PR removes the creation of temporary files for the following:

* component map (`.idyll/components.js`)
* data map (`.idyll/data.js`)
* AST JSON (`.idyll/ast.json`)
* syntax highlighter registration (`.idyll/syntax-highlighters.js `)

Instead it passes streams directly to browserify. The files (now streams) are update to prefer absolute paths rather than relative ones, since they don't exist in a directory anymore and just absolute seems like less logic to worry about.


mathisonian:
I need to figure out what is going on with the tests

src/pipeline/bundle-js.js
```diff
@@ -31,12 +41,29 @@ module.exports = function (opts, paths) {
...
+            requireName: '__IDYLL_DATA__'
+          }, {
+            key: 'components',
+            requireName: '__IDYLL_COMPONENTS__'
+          }, {
+            key: 'ast',
+            requireName: '__IDYLL_AST__'
+          }];
+          requires.forEach((obj) => {
+            b.exclude(obj.requireName);
```
mathisonian: this `exclude` call is necessary because of a browserify bug

src/pipeline/bundle-js.js
```diff
@@ -31,12 +41,29 @@ module.exports = function (opts, paths) {
...
+
+            b.require(toStream(output[obj.key]), {
```
mathisonian: there's probably a way to create the input as streams directly instead of creating strings and converting, but this seems fine for the moment

src/pipeline/parse.js
```diff
@@ -144,13 +138,12 @@ exports.getHighlightJS = (ast, paths) => {
...
-  let js = `const rsh = require('${slash(path.join(rshPath, 'light'))}');`
-
+  let js = `var rsh = require('react-syntax-highlighter/dist/light');`
   Object.keys(languages).forEach((language) => {
     if (languageMap[language]) {
       language = languageMap[language];
     }
-    js += `\nrsh.registerLanguage('${language}', require('${slash(path.join(rshPath, 'languages', language))}').default);`
+    js += `\nrsh.registerLanguage('${language}', require('react-syntax-highlighter/dist/languages/${language}').default);`
   });
```
mathisonian: I removed the logic here temporarily, it turns out what we have implemented currently is still broken under certain conditions 😞

mathisonian: The current implementation is broken if idyll is linked and a `defaultComponent` folder is used. @bclinkinbeard can you remind me in which scenario it wasn't working to have this just `require('react-syntax-highlighter')` instead of the explicit path?

One thing I did notice in my dive into browserify docs is there is a `paths` option that we can use to expand which paths browserify will use to look for modules, which may be useful for cases like this.

bclinkinbeard:
@mathisonian hope you don't mind I did a bit of cleanup that I noticed while getting my head around these changes. Looking good though! Tests are passing locally and this should be helpful for #75.

src/pipeline/parse.js
bclinkinbeard: @mathisonian I think `require('react-syntax-highlighter')` would fail any time the input project doesn't list it as a dependency.

mathisonian:
@bclinkinbeard changes look great! I just pushed a fix that should allow the `react-syntax-highlighter` to work properly under all the different conditions.

src/pipeline/bundle-js.js
```diff
@@ -39,6 +39,10 @@ module.exports = function (opts, paths, output) {
       fullPaths: true,
       cacheFile: path.join(paths.TMP_DIR, 'browserify-cache.json'),
       transform: getTransform(opts),
+      paths: [
+        path.join(paths.INPUT_DIR, 'node_modules'),
```
bclinkinbeard: @mathisonian will this cause an issue if the input dir doesn't have a node_modules folder?

PR #79, opened 2017-06-07
fix subtle difference between server and client syntax highlighting

mathisonian:
This makes it so `react-syntax-highlighter` is required in the exact same way on the server and the client.

I'm open to ways to clean up the implementation a bit, but this seems to get the job done

### DIFF
--- a/src/pipeline/parse.js
+++ b/src/pipeline/parse.js
@@ -111,7 +111,7 @@ exports.getDataJS = (ast, DATA_DIR, o) => {
 }
 
 
-exports.getHighlightJS = (ast, paths) => {
+exports.getHighlightJS = (ast, paths, server) => {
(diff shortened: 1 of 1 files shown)


src/pipeline/parse.js
```diff
@@ -172,7 +188,7 @@ exports.getHTML = (paths, ast, components, datasets, template) => {
     componentClasses[key] = require(components[key])
   })
 
-  require(resolve.sync('react-syntax-highlighter', { basedir: paths.DEFAULT_COMPONENTS_DIR }));
+  requireHighlightJS(ast, paths);
```
bclinkinbeard: Can we just do `exports.getHighlightJS(ast, paths, true);` directly here and get rid of `requireHighlightJS`?

Issue #88, opened 2017-06-11
Consider putting the InteractiveDocument component into its own module


mathisonian:
This remove most of the client side code from this module, which I think probably makes sense. It would also be nice to be testing the component independently to make sure the dataflow, ast parsing, etc. works as expected

Issue #51, opened 2017-05-11
Investigate separate bundles


bclinkinbeard:
The files required in `components.js` will rarely change during normal content development workflows. I *think* we should be able to bundle them on their own, and mark them as external in the main bundle, and get much faster bundle times.

PR #91, opened 2017-06-13
Faster! Faster!

bclinkinbeard:
This commit skips SSR when minify is turned off, which saves about 500ms from the build time. (Interestingly enough, the time seems to be spent requiring the component files, not doing the actual render to string.) I'm not sure tying it to minify (or watch) is the best way to determine the behavior, so maybe we should add a `--ssr` flag? I think if `watch` is true you definitely don't want/need SSR, but for something like a REPL `watch` would be false but you still don't want SSR.

Additionally, it appears we can shave about 300ms from the bundle time by setting `react` and `react-dom` as external. I can consistently get build times under 400ms in that scenario so I think it's worth doing. The question is how do we include those dependencies then? Maybe that's also conditional and we stick a couple of script tags in the HTML in that case?

### DIFF
--- a/bin/idyll.js
+++ b/bin/idyll.js
@@ -16,7 +16,7 @@ var argv = require('yargs')
     l: 'layout',
     n: 'no-minify',
     o: 'output',
-    k: 'spellcheck',
+    k: 'no-spellcheck',
     r: 'no-ssr',
     t: 'template',
     e: 'theme',
@@ -36,9 +36,8 @@ var argv = require('yargs')
   .describe('no-minify', 'Skip JS minification')
   .describe('output', 'Directory where built files should be written')
   .default('output', 'build')
-  .boolean('spellcheck')
-  .default('spellcheck', true)
-  .describe('spellcheck', 'Check spelling of Idyll source input')
+  .boolean('no-spellcheck')
+  .describe('no-spellcheck', `Don't check spelling of Idyll source input`)
   .boolean('no-ssr')
   .describe('no-ssr', 'Do not pre-render HTML as part of the build')
   .describe('template', 'Path to HTML template')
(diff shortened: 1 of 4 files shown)


bclinkinbeard:
Related/similar to #51 (and #88) maybe we should create a separate bundle that contains React, React DOM, and the set of default components. It could be built ahead of time, maybe as part of the `idyll-default-components` publish process? That would ensure that Idyll only has to parse and build the document and any custom components being used.

Thoughts?

mathisonian:
@bclinkinbeard if we built a bundle we would we still be able to only include code for the components actually being used? i'd like to keep the bundle sizes small if possible. if not just offloading `react` and `react-dom` might make sense for now

bclinkinbeard:
@mathisonian Good question. Probably not, unless we implemented a build that could do tree shaking. That's obviously a bigger undertaking so you're probably right about sticking with the react bits for now.

What are your thoughts on a `--ssr` (or similar) flag?

mathisonian:
I think an SSR flag makes sense as opposed to tying it to `minify`. I'm inclined to think it should be enabled by default if `minify` is enabled by default as well?

I think we should make sure there is reasonable error handling and a message printed to console in case SSR failed as I could see that being a point of frustration for users if there build keeps failing because some component doesn't work exactly right with ssr.

bclinkinbeard:
Added `--no-ssr` option and converted `--spellcheck` to `--no-spellcheck` for consistency. Now all the options that default to `true` are turned off via options beginning with `no`.

In terms of error handling and communicating errors I agree but would prefer to handle that in a separate PR since this one is enabling SSR to be turned off, not making it run when it previously wouldn't.

I guess we can defer the external stuff to another PR as well, because it's going to be a large effort. I toyed with Rollup and Webpack this morning to see how hard tree shaking might be and the answer appears to be very. I think this change is worth publishing on its own and then we can revisit the other aspects separately.

bclinkinbeard:
I made some progress on externalizing React and friends from bundles but definitely don't want to tackle it in this PR. @mathisonian if the code here looks OK to you go ahead and push the green button. :)

mathisonian:
This looks good, i think your note about handling errors + externs in separate PRs makes sense 👍

PR #92, opened 2017-06-19
Conditionally exclude react from project bundles

bclinkinbeard:
I'm not sure if a new dev flag is the best way to toggle this so I'm open to suggestions on that. I'm also not sure if we'd ever want to enable this alongside SSR, but that is not currently supported.

This does work though, and I can consistently get builds in the 250ms to 350ms range using it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,7 @@ jspm_packages
 # Optional REPL history
 .node_repl_history
 
-.ast.json
+src/client/assets/libs.js
 
 !test/**/node_modules
 test/*/src/.idyll
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "idyll": "bin/idyll.js"
   },
   "scripts": {
+    "prepublish": "node scripts/build-assets.js",
     "test": "jest"
   },
   "repository": {
--- a/scripts/build-assets.js
+++ b/scripts/build-assets.js
@@ -1,0 +1,30 @@
+const fs = require('fs');
+const { join } = require('path');
+const b = require('browserify')();
+
+b.require(join(
+  __dirname,
+  '..',
+  'node_modules',
+  'react',
+  'dist',
+  'react.min.js'
+), { expose: 'react' });
+
+b.require(join(
+  __dirname,
+  '..',
+  'node_modules',
+  'react-dom',
+  'dist',
+  'react-dom.min.js'
+), { expose: 'react-dom' });
+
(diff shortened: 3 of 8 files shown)


mathisonian:
Thanks for the PR @bclinkinbeard! For thinking about the dev flag vs. other options of enabling, are there any cases where this would be wanted besides for pure performance reasons during dev?

Is that 250-350ms down from like 500?

bclinkinbeard:
The only other reason I can think of is swapping out React or something like that, but I'd say it's probably just performance. I still need to investigate but I seemed to be getting much slower builds on Heroku than I do locally, so the host environment may be a factor too. I just don't want to get too crazy with the number of options since we already have several.

It seems to shave about 200ms off the build time. Here are the logs in watch mode:

```
Build Time: 838.178ms
Build Time: 632.488ms
Build Time: 432.186ms
Build Time: 432.198ms
Build Time: 408.185ms
Build Time: 487.205ms

With --dev flag:
Build Time: 546.271ms
Build Time: 373.157ms
Build Time: 269.498ms
Build Time: 253.164ms
Build Time: 484.479ms
Build Time: 349.544ms
Build Time: 290.470ms
```

And without watch mode:
```
Build Time: 816.022ms
Build Time: 783.828ms
Build Time: 846.649ms
Build Time: 804.963ms
Build Time: 785.149ms
Build Time: 807.045ms

With --dev flag:
Build Time: 564.214ms
Build Time: 557.695ms
Build Time: 572.520ms
Build Time: 551.557ms
Build Time: 537.639ms
Build Time: 546.774ms
```

mathisonian:
To avoid more complication in the flags maybe we should just automatically do it when `watch` is true since [there is a precedent for this](https://github.com/idyll-lang/idyll/blob/master/src/index.js#L41)? 

To enable it when `watch` is false, we could either give it its own opt-in flag, or have a flag that encapsulates things like this and minification  (and potentially others)

so 

```
idyll index.idl --minify --external
```

 vs.

```
idyll index.idl --fast-build
```

I'm a little hesitant about re-introducing javascript written to the actual fs during watch, but maybe that's okay.

bclinkinbeard:
Holding off on this while I grow my sample set and focus on other things. Build time seems to be somewhat unpredictable and I'd like to avoid adding this if possible.

bclinkinbeard:
Closing since this is way out of date and not a big concern right now. Can revisit down the road if necessary.

PR #98, opened 2017-06-27
only transform necessary front-end files

mathisonian:
This reduces the number of files that get registered with babel to only include those files in component folders. It fixes a bug where the registering the files would cause issues in other programs that `require('idyll')`.

/cc @bclinkinbeard


bclinkinbeard:
LGTM, nice work!

PR #101, opened 2017-07-01
remove need for special updateProps on idyll component

mathisonian:
This refactors the `nodeWalker` to dynamically inject the special `updateProps` method, instead of relying on components to extend a special component that implements them. 

This also adds `memoizee` to cut down on repeated work in the node walker across state updates, making things slightly faster. There is probably some more work to be done here to make things as fast as possible, but thats for a different PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "idyll-compiler": "^1.4.4",
     "idyll-default-components": "^1.3.2",
     "insert-css": "^2.0.0",
+    "memoizee": "^0.4.5",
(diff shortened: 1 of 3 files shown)


mathisonian:
This should be merged along with https://github.com/idyll-lang/idyll-component/pull/5/

bclinkinbeard:
I only have a vague understanding of this code, but I don't see any issues with it. It would be nice to get rid of the console warnings. :)

PR #104, opened 2017-07-11
use separate package for client-side component

### LINKED ISSUE #88: Consider putting the InteractiveDocument component into its own module

mathisonian:
This will make it easier to test that component, and separates the logic a bit more cleanly. This fixes #88. 

It will also be good to keep issues for that component in their own tracker since that logic is fairly complex.


bclinkinbeard:
I like the separation but wonder if we really want another package. I'm fine with a new package if you are but maybe it's time to consider a monorepo?

mathisonian:
I think I'm okay with adding another repo, as these are conceptional very different pieces. 

I'm not opposed to mono-repo at all, but don't really have familiarity with lerna (assuming we'd need something like that), getting the CI scripts to work etc. Let's go ahead with this for now and have a discussion about merging everything separately.

PR #120, opened 2017-08-11
WIP: Idyll doc updates

bclinkinbeard:
This PR aims to clean up the document creation to allow for better debugging and future evolution. After some housekeeping I added tests for the existing behavior that verified, given an AST, the doc's `state` and `derivedVars` are populated as expected. With the tests in place I implemented new logic for parsing the AST into those properties.

(Side note, the `getNodesByName` function is essentially borrowed from https://github.com/idyll-lang/idyll/blob/master/packages/idyll-cli/src/pipeline/parse.js#L17 so we may want to extract some common utils at some point.)

The new approach has one big difference, as seen in 4ec9a62. The existing implementation of `update` on the derived vars relied on a lot of global state that we want to avoid. The new code relies on `updateDerivedVars`, defined on `InteractiveDocument`, and `this.derivedVars.foo.update(newState)` is no longer supported. I don't think this will cause any issues but @mathisonian can confirm.

I am now moving on to the datasets and refs so we can get rid of the var walker entirely.

Feedback please!


mathisonian:
This looks great. I'll need to take a closer look at the derived vars. I don't think this should cause issues, but we were running into some subtle things before with getting the state out of sync there.

bclinkinbeard:
OK, this is in pretty good shape now. Most everything has been ported to the new structure using the JSON to React elements utility and other pure functions. The only things I haven't ported yet are the refs stuff and what looks to be special handling of props beginning with `on` or `handle`. @mathisonian if you can shed any light on those that'd be cool.

Next Thursday I will beef up the tests on the new stuff, though they're decent already. Besides that I'll also start working on some of the component "primitives" we talked about.

bclinkinbeard:
@mathisonian 0e7ed85 implements smart updates so the only components that re-render are the ones that need to!

mathisonian:
Really clean - great work @bclinkinbeard

bclinkinbeard:
@mathisonian 1afc267 reworks the refs calculations. It uses https://github.com/stutrek/scrollMonitor, which handles the relative scrolling stuff (supporting containers as well as window) and gives some extra props like `isFullyInViewport`, `isAboveViewport`, etc.

The code does not currently implement the `scrollProgress` value. I'd like to talk through that feature because I'm not sure the existing setup provides sufficient value. Rather than a 0 to 1 value I wonder if maybe a `percentVisible` prop would be helpful. For example, on something like https://idyll-lang.github.io/idyll/scroll/ maybe you could configure it to trigger a change when each "screen" is >= 50% visible. Also, combined with `isAboveViewport` and `isBelowViewport` you could discern exactly where something is.

I'm going to start working on propagating refs updates but would love your feedback on the above.

mathisonian:
Cool, using scrollMonitor makes sense - my only issue with it is we already have a dependency on https://edull24.github.io/ScrollWatch/ in `IdyllComponent`, I wonder if we can consolidate to just depend on one of them. 

I'm wondering what it makes sense to have baked into the refs vs what should be delegated to a wrapping scroll component. I'm working on the API for a scroll component based on some of what @rreusser has done (e.g. http://rickyreusser.com/idyll-experiments/fixed-feature/), I think seeing that might help inform some of the decisions here. I should have something posted on that in the next couple of days.

bclinkinbeard:
Yea we can actually get rid of `IdyllComponent` entirely. This PR makes it so all components get an `updateProps` prop added dynamically, so there's no need to extend a base component. (See line 233 of index.js)

I agree on waiting and seeing what we can delegate to components.

bclinkinbeard:
Any additional thoughts on this @mathisonian? (I realize it's a huge amount of changes.) I'd like to get this merged this week if possible.

mathisonian:
I've set aside all day tomorrow to do thorough testing and any cleanup. Hopefully can get it merged then

mathisonian:
Hey @bclinkinbeard, this is looking really good. I think I've wrapped my head around the changes and the mostly make sense. I've been testing locally in the editor and am getting one error I'm wondering if you can reproduce: 

The steps to reproduce are just to change the equation in

```
[Chart
  equation:`(t) => Math.sin(t * frequency)`
  domain:`[0, 2 * Math.PI]`
  samplePoints:100 /]
```

to 

```
[Chart
  equation:`(t) => Math.cos(t * frequency)`
  domain:`[0, 2 * Math.PI]`
  samplePoints:100 /]
``` 

and move the slider below the chart. It gives a fairly cryptic react error for me - 

```
Uncaught TypeError: Cannot read property '_currentElement' of null
    at ReactCompositeComponentWrapper._updateRenderedComponent (index.js:111568)
    at ReactCompositeComponentWrapper._performComponentUpdate (index.js:111547)
    at ReactCompositeComponentWrapper.updateComponent (index.js:111468)
    at ReactCompositeComponentWrapper.receiveComponent (index.js:111370)
    at Object.receiveComponent (index.js:116509)
    at Object.updateChildren (index.js:110728)
    at ReactDOMComponent._reconcilerUpdateChildren (index.js:115792)
    at ReactDOMComponent._updateChildren (index.js:115896)
    at ReactDOMComponent.updateChildren (index.js:115883)
    at ReactDOMComponent._updateDOMChildren (index.js:112783)
```

I also pushed a small change removing some old imports - hopefully those were just leftover from before. 

One other thing - the change from `this.updateProps` to `this.props.updateProps` feels a little wrong semantically. Is there any way to avoid attaching this function to the props?

mathisonian:
It also seems like the scroll code may be triggering a some extra repaints that weren't there before - see attached video (green == repaint)

old
repaint on scroll - old

new
repaint on scroll

in particular it seems like maybe code blocks are getting repainted on scroll for some reason

mathisonian:
I think there's a bug in the refs logic where ``[Display var:`refs.scatterChart.position.top` /]`` works as expected, but ``[Display var:`1 + refs.scatterChart.position.top` /]`` does not

bclinkinbeard:
Cool, I will check the errors and extra paints and report back.

re: `updateProps` I actually think attaching it to `props` is more semantically correct. :) `props` is the standard mechanism for passing data and functionality to another component as seen in general dev and libs like `react-router`. When a component is used to render a route it gets `match`, `location`, and `history` props. `updateProps` feels similar because it's not part of the original component definition but something the component is "decorated" with when running in Idyll.

I do wonder if a more explicit name would be better. Something like `updateIdyll` or `updateIdyllProps` or `setIdyllState`?

bclinkinbeard:
I'm somewhat stumped on the `Math.cos` thing. It looks like it's just getting into a bad state from the errors while typing (`Math.co is not a function`, etc). The weird thing is the existing, deployed editor throws the same errors, but it doesn't put things into a bad state. I'm also confused about why those errors are being logged in the deployed version as that code should be wrapped in a try catch. I will dig in further tomorrow but my initial attempts at try catchery did not help.

bclinkinbeard:
What are you using to examine repaints? I've been checking with the React Dev Tool's Highlight Updates turned on and it's not showing the same results as your recording. (Not doubting the results, just want to repro.)

mathisonian:
I'm using the chrome devtools -> rendering -> show paint and FPS tools

I can help dig into the state issue as well

bclinkinbeard:
@mathisonian I just pushed a couple commits that fix the "refs in expressions" problem you found. My initial implementation was just naïve. Should be good to go now, and I included a few in the `initial.idl` for easy testing. We'll obviously overhaul that initial content.

bclinkinbeard:
I'm seeing the entire window repaint on scroll, even on the deployed version.

repaints

bclinkinbeard:
[This thread](https://github.com/facebook/react/issues/6895#issuecomment-281405036) explains that the error we're seeing is caused by uncaught exceptions in component render functions. Sure enough, if the call to `this.props.equation` in the Chart component is wrapped in a try...catch, the error is avoided. That's obviously not a burden we want to put on authors.

If I wrap the calls to `evalExpression` in a try catch it keeps the app from breaking, but it still logs the `_currentElement` error on subsequent renders. I have not yet figured out how to prevent that error like the deployed version seems to do.

mathisonian:
This seems like something that the React 16 error boundaries would help with - although, I don't quite understand why the old and new versions would be showing different behaviors here

mathisonian:
I can work with `updateProps` being passed in as a property. Changing to `updateIdyll` might be a little more clear but lets save it for a different discussion/PR since that will mean documentation updates too.

mathisonian:
I think this is looking good. Anything else you want to update before we merge? 

We'll need to revert some of the debugging changes and should figure out the best time to publish the modules under their new names.

bclinkinbeard:
Sweet! I will get rid of the blue highlight (I want to enable this to happen conditionally in a future PR) and put `initial.idl` back to a reasonable state.

Once I do that I will merge and start integrating with my app locally. I'll be back on this project next Thursday and Friday, maybe we can publish then, with discussion and minor clean up in the interim?

Issue #124, opened 2017-08-12
Minification errors


mathisonian:
We've had multiple issues with code minification breaking prop updates in the current release. @RERossel reported this in the gitter, and @JobLeonard was seeing the same thing in #117

mathisonian:
<img width="1280" alt="screen shot 2017-08-12 at 3 07 36 pm" src="https://user-images.githubusercontent.com/1074773/29244578-25d9e7ae-7f70-11e7-9a0c-a73dfc5a8c1b.png">

this error is happing in the default project after `npm run build`

bclinkinbeard:
Did uglify start mangling names where it previously didn't?

________________________________

mathisonian:
I'm looking into it now, seems like either that or something with the new method of injecting that property 😠

JobLeonard:
You know, I just noticed that in my code we wrap a `SketchComponent extends IdyllComponent` in a `Sketch extends Component`, and the latter passes down `updateProps`. However, only the latter is actually used as a component in the `.idl` file. Hold on.

... yep, changing that to `Sketch extends IdyllComponent` seems to fix it, which makes sense I guess. But that begs the question of why it _did_ work fine (when it shouldn't) in development mode

mathisonian:
@JobLeonard 

We just switched the technique we were using to inject the `updateProps` property dynamically so that you could use a standard react component and not necessarily need to extend from `IdyllComponent`. So it should still work just extending `Component`, but it seems like this change isn't interacting well with uglify.

mathisonian:
Alright, I've found the propblem. In [this code](https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/src/visitors/node.js#L44-L53) we have a conditional that depends on the results of `isClass()` a check to typecheck if something is a JavaScript class or not. This check works fine without minification but fails if the code is minified. We are using the `is-class` module for this https://github.com/miguelmota/is-class.

mathisonian:
Fix submitted in #126

PR #126, opened 2017-08-12
fix class check in minified code

mathisonian:
This fixes https://github.com/idyll-lang/idyll/issues/124 by updating uglify to leave the functions intact enough for our class check to still work.


bclinkinbeard:
LGTM

PR #145, opened 2017-09-07
error boundaries

mathisonian:
submitting more for consideration / discussion than to merge straightaway. this allows us to do some nice handling of runtime errors in components since we already wrap them all 

errror

### DIFF
--- a/packages/idyll-components/components/gist.js
+++ b/packages/idyll-components/components/gist.js
@@ -1,5 +1,6 @@
 const React = require('react');
 const IdyllComponent = require('idyll-component');
+const PropTypes = require('prop-types');
 
 class EmbeddedGist extends IdyllComponent {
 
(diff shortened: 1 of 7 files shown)


bclinkinbeard:
This looks killer!!!

bclinkinbeard:
I just had an idea. What if we checked for a prop on the wrapped component called `errorHelper` or something and, if it was defined, displayed it below the error? Component authors (especially us) could then define something to be displayed when the component has an error. It could list and describe supported props, show example syntax, etc. I think that could be super helpful.

mathisonian:
I love that

bclinkinbeard:
We'll need to have `IdyllDocument` itself catch errors as well. Currently if `rjs.parseSchema(transformedSchema)` blows up due to a missing component or something the whole document is hosed.

Kind of tempted to just upgrade to React 16 now and put this stuff in. We only have a few enzyme tests anyway and we could update them as part of a larger testing push when v3 lands. Thoughts @mathisonian?

bclinkinbeard:
Closing in favor of #170 since the base implementation from this PR has already been done

PR #149, opened 2017-09-09
Add scroll components

mathisonian:
Adds the following components to help setup scrolling stories:

*Feature* - A "fixed" scrolling feature. This component will stick to the top of the page as you scroll through its contents
  - Content - The fixed content of the Feature component

```
[var name:"index" value:1 /]

[Feature]
  [FeatureContent]
    [Component value:index /]
  [/FeatureContent]

  [Waypoint onEnteredView:`index = 1`] First Waypoint[/Waypoint]
  [Waypoint onEnteredView:`index = 2`] Second Waypoint[/Waypoint]
  [Waypoint onEnteredView:`index = 3`] Third Waypoint[/Waypoint]
[/Feature]
```

*Full Screen* - a component that takes up the entire screen, good for full-bleed images

```
[FullScreen backgroundImage:`"images/my-fullscreen-image.jpg"`  /]
```

*Panel* - if used after a `Feature` component, the panel will cause the following content to slide up over it (see attached gif)

*Waypoint* A component used to trigger events on scroll 

Also adds a helper to make it easy to preload images: 

```
[Preload images:`["image1.jpg", "image2.jpg", ...]` /]
```

I'll need to add some CSS to this before it gets merged 

scrolling-optimized

that gif is me trying to replicate the techniques in http://nation.maps.arcgis.com/apps/Cascade/index.html?appid=de7f932e3a1d494f9c9d9a67fb0de646


bclinkinbeard:
The demo looks awesome! Does `Feature.Content` work? I had forgotten to mention it but I thought that was something we lost in the refactor.

Though now that I'm thinking about it, that resolution happens in the CLI, doesn't it? We might need to review the CLI for compatibility with the editor. I don't think we'd want to have things that work in one context but not the other.

mathisonian:
I was testing this against the latest NPM release - it was working there. Good point, we should test against the new changes and make sure things aren't out of sync

mathisonian:
I've updated this to not use the Idyll component, correctly use the new `props.updateProps`, and not depend on nested dot notation. 

It should be good to merge with the current state of things.

bclinkinbeard:
These aren't exposed on the index and I'm not sure how you want them to be.

PR #150, opened 2017-09-10
Remove idyll component and implement hook functions

bclinkinbeard:
This adds support for hooks to all (wrapped) components. Currently implemented hooks are `onScroll`, `onEnterView`, and `onExitView`, though others will be easy to add if we want. Note that the enter hook is implemented to fire when a component comes into view even partially but exit only fires when a component fully leaves view. We probably want to refine the list of hooks to something like this:

- `onEnterViewPartially`
- `onEnterViewFully`
- `onExitViewPartially`
- `onExitViewFully`

I also thought about things like adding a `percentInView` property to the ref objects, or supporting hooks like `in20`, `out70`, etc. but I'd like to know what you think. We'd obviously need to be careful of performance issues if we go too far beyond the above list.

The last commit in this PR adds hooks to the scatter chart for easy testing. I'll back that out before merge.


packages/idyll-document/src/index.js
```diff
@@ -187,19 +212,21 @@ class IdyllDocument extends React.PureComponent {
     const transformedSchema = mapTree(
       schema,
       node => {
+        if (typeof node === 'string') return node;
+
         // transform refs from strings to functions and store them
-        if (typeof node !== 'string' && node.ref) {
-          const refName = node.ref;
+        if (node.ref || node.hasHook) {
+          node.refName = node.ref || node.component + Math.round(Math.random() * 99999).toString();
```
mathisonian: Can we change this to something deterministic? Maybe just an incrementing counter

mathisonian:
I think allowing the user to specify a percent in view seems like a reasonable approach. I'm not sure i follow why this would lead to perf issues... shouldn't the performance just be effected by how many handlers are placed on the page?

mathisonian:
Perhaps @tonyhschu would have some input as to the preferred API for scroll callbacks as a new user of Idyll. 

@tonyhschu for context, we've revamped a lot of the internals and taking the opportunity to think about what new properties or features should be supported to better facilitate scroll based logic. We've also created some new helper components in https://github.com/idyll-lang/idyll/pull/149

packages/idyll-document/src/index.js
bclinkinbeard: Sure

bclinkinbeard:
Regarding performance concerns, I guess the real determining factor would be how many refs are being tracked. `percentInView` is definitely an easy, cheap calculation. 

Now that I'm thinking about it, the difficulty with the `in20` kind of hook probably isn't performance, but knowing when to call it. Like, does it get called any time `Math.round(percentInView) === 20`? I suppose that's easy enough but now I'm starting to think maybe we keep those kinds of things out of the core for now. It would be easy enough to implement as a custom component anyhow, and adding hooks would only require a minor release.

What if we do the ones below for the initial release, and let use/discussion dictate what we add afterwards?

- `onEnterViewPartially`
- `onEnterViewFully`
- `onExitViewPartially`
- `onExitViewFully`
- `percentInView`

tonyhschu:
More thoughts to come, but responding to Ben's immediate comment above with an example that might help solidify a use case.

The way I've handle something like `in20` in the past is using RxJS's `distinctUntilChanged` functionality. The use case would often be something like this.

```css
.annotation {
  transition: opacity 0.3s, transform 0.2s;
  opacity: 0;
  transform: translate(0, -10px);
}
.annotation.show {
  opacity: 1;
  transform: translate(0, 0);
}
```

At some boundary, e.g. 20% down a section, I want to add `.show` to an annotation element. It would then fade in and slide in. As soon as I scroll above 20%, it should fade out and slide up.

In React, I would basically have a `className` that set like this

```js
className={(isPast20) ? "annotation show" : "annotation"}
```

Being able to specify that a value depends on a boundary is useful for all sorts of CSS animations.

bclinkinbeard:
That's interesting @tonyhschu, thanks. So you're picturing something like this on the ref object?

```
in10: percentInView >= 10,
in20: percentInView >= 20,
in30: percentInView >= 30,
in40: percentInView >= 40,
in50: percentInView >= 50,
in60: percentInView >= 60,
in70: percentInView >= 70,
in80: percentInView >= 80,
in90: percentInView >= 90,
```

tonyhschu:
Hmm. I didn't think of it in context of a refs object.. maybe? A thing that I did try to do, but couldn't get to work was using derived variables for this purpose, e.g.

```
[derived name:"shouldShow" value:`refs.sectionName.scrollProgress.y >= 0.1` /]
```

Then using `shouldShow` elsewhere to control several different CSS animations. (It may not have worked because I simply got the syntax wrong... hopefully my intent is clear here.) I haven't dug into the code base to see how derived variables are implemented, but my mental model was that it worked like RxJS observables, which is why I tried creating that.

bclinkinbeard:
The ref object is just how we store all the info for a given node. You can see the code where we construct them [here](https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/src/index.js#L27).

I _think_ you should be able to create a derived var like that but we don't currently have `scrollProgress` on the ref objects (hence this discussion).

One thing to note is the distinction between refs in general and the hooks we've talked about in this thread. Hooks are meant to allow a component to perform some action based on their own position, and will have the doc state and their own ref object passed in. 

```
[Chart type:"scatter" onEnterView:`doSomething()`]
...
[Chart type:"line" in30:`doSomethingNowThatImThirtyPctVisible()`]
```

Things that reference another ref, like your derived var, could just use the `percentInView` we mentioned.

```
[Derived name:"shouldShow" value:`refs.sectionName.percentInView >= 30` /]
```

mathisonian:
@bclinkinbeard I think that list you have above makes sense for this PR, and the derived variable seems like a reasonable way to observe these boundaries, rather than adding a bunch of % based events. 

An alternative suggestion would be to add a second property that defines the `percentInView` you care about 

``[Chart onEnterView:`...` enterViewThreshold:0.3 /]``, ``[Chart onEnterView:`...` enterViewThreshold:-.1 /]``, etc

For names, how about - 

- `onEnterView`
- `onEnterViewFully`
- `onExitView`
- `onExitViewFully`

For completeness I think its nice to have both x and y covered in the `percentInView` parameter in the way that we have that in `scrollProgress` now. 

One important point that @tonyhschu brought up at the bottom of https://tonyhschu.github.io/idyll-scroll-linked/ is that its useful for this value to not be bounded from [0, 1] (or [0, 100] if thats how we decide to implement it - personally i'd think [0, 1] is less surprising)

mathisonian:
In the interest of landing this - I'd be fine to get just these events in and add the refs property back in another PR. 

I'm curious what people think about adding the `threshold` props as this seems like a nice way to get the desired functionality to avoid having to use derived variables or extra listeners

bclinkinbeard:
I like the idea of a threshold prop but don't understand "useful for this value to not be bounded from [0, 1]". Is there a use case not covered by the four methods @mathisonian mentioned plus a threshold between 0 and 1? We could even only call the method when the threshold is met.

mathisonian:
A main use case would be "trigger event X slightly before content Y" comes into view

bclinkinbeard:
[-1, 2]?

mathisonian:
Yeah, that could work. I guess it depends on how we implement it, but I'm not sure there's an advantage to having something bounded vs having something that goes (-∞,∞). 

Where 
* negative value means you haven't reached the content yet,  
* [0, 1] means the content is on the screen, and 
* something > 1 means you've scrolled past it. 

the only issue with this is that it isn't really representing `percentInView`, its something slightly more abstract

bclinkinbeard:
Actually, I think we should let users do that stuff at the component level using the ref object. I think the four hooks and percentInView is a reasonable starting point. We can always add to it but I'd prefer to start minimal.

mathisonian:
Sounds good 👍

bclinkinbeard:
I was about to commit and push but now I'm wondering: is there a need for `onEnterViewFully` and `onExitViewFully` since the ref object has `isFullyInViewport` on it? And we're adding `percentInView` which could be checked too?

mathisonian:
To replicate it would you add an `onScroll` listener and then check those variables? I think they'd still be nice shortcuts to have if they're straightforward to add

bclinkinbeard:
Ah, yea, I forgot to ask about `onScroll`. I assume we want to keep that? I'll go ahead and push what I have and we can discuss.

mathisonian:
Will take a closer look but, at overall this looks great to me. Makes sense to me to keep `onScroll` and 🎉 for removing `IdyllComponent`.

bclinkinbeard:
Looks like just the CLI tests are failing. Do we want to fix that here or in a separate PR?

mathisonian:
That should be fixed once #159 is merged. I think this is good to go

bclinkinbeard:
Woohoo! Clicking the button...

bclinkinbeard:
Ugh, just realized I forgot to add `percentInView`.

Issue #152, opened 2017-09-11
Getting CLI + new idyll-doc to work together


mathisonian:
I'm seeing some issues when using the CLI with the new document code - in particular the follow error which looks like the component lookup isn't happening exactly as expected:

```
Error: ReactJsonSchema could not find an implementation for: CodeHighlight
    at ReactJsonSchema.resolveComponent (/Users/mathisonian/projects/idyll/idyll/packages/idyll-document/dist/utils/schema2element.js:102:17)
    at ReactJsonSchema.createComponent (/Users/mathisonian/projects/idyll/idyll/packages/idyll-document/dist/utils/schema2element.js:82:28)
    at ReactJsonSchema.parseSchema (/Users/mathisonian/projects/idyll/idyll/packages/idyll-document/dist/utils/schema2element.js:30:24)
    at ReactJsonSchema.parseSubSchemas (/Users/mathisonian/projects/idyll/idyll/packages/idyll-document/dist/utils/schema2element.js:53:34)
```

where the `componentMap` looks like 
```
{ header: [Function: Header],
  range: { [Function: Range] defaultProps: { value: 0, min: 0, max: 1, step: 1 } },
  display: [Function: Display],
  'code-highlight': [Function: CodeHighlight],
  'custom-component': [Function: CustomComponent],
  Wrapper: [Function: Wrapper] }
```

with the relevant code here https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/src/utils/schema2element.js#L44-L54

PR #153, opened 2017-09-11
add param case

mathisonian:
Re: https://github.com/idyll-lang/idyll/issues/152, I'm trying to make sure things work out of the box with the CLI and the new `IdyllDocument`. This has made one change so far:

1. Add another lookup case of component names to ensure the component map passed from the CLI works correctly.

### DIFF
--- a/packages/idyll-document/src/index.js
+++ b/packages/idyll-document/src/index.js
@@ -120,7 +120,8 @@ class Wrapper extends React.PureComponent {
 
   onUpdateRefs(newState) {
     const nextState = {};
-    Object.entries(this.props.__expr__).forEach(([key, val]) => {
+    Object.keys(this.props.__expr__).forEach((key) => {
+      const val = this.props.__expr__[key];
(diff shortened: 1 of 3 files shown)


bclinkinbeard:
Do we really need to worry about Node 6? Current version is 8.4.

We're also using `Object.entries` in a couple places which has the same support. I vote for a "current plus previous major" approach.

mathisonian:
I'd prefer to go at least 2 back since it is user facing, and in my experience a lot of people are slow to upgrade or just use the even-numbered LTS versions. 

I don't mind doing the PRs to fix any existing issues

bclinkinbeard:
Can we put the changes in their own PR and https://www.npmjs.com/package/object.values and https://www.npmjs.com/package/object.entries? That way we can just remove the shims at some point in the future.

mathisonian:
Okay, I've reverted those changes here and opened https://github.com/idyll-lang/idyll/pull/156 with the shims

PR #158, opened 2017-09-13
Working to re-enable CI

mathisonian:
Travis:
- [ ] idyll-cli
- [x] idyll-compiler
- [x] idyll-components
- [x] idyll-document

I've added an initial appveyor script, but I think we'll have to merge it into master before appveyor starts picking up changes in the repo again, so we can merge that part as is and deal with the details in a different PR


mathisonian:
screen shot 2017-09-12 at 7 00 27 pm

Progress so far - 

it looks like the issues are 
- `idyll-components` doesn't include the dist files, so when `idyll-document` test runs, this fails with `idyll-components` not found
- `idyll-compiler` doesn't include the generated JS grammar, so when the `idyll-cli` tries to invoke the compiler it fails

mathisonian:
screen shot 2017-09-12 at 7 46 19 pm

Making progress - 

- idyll-document is failing on node v6 which is expected before https://github.com/idyll-lang/idyll/pull/156 is merged
- idyll-cli is failing due to my change to the way that the compiler wraps content in divs, but this is to be expected

@bclinkinbeard one thing before merging - `idyll-document` has react and react-dom listed as both dev and peer dependencies, i guess we should just choose one or the other

bclinkinbeard:
Having both peer and dev deps was actually intentional so you have what you need when developing. 😀

PR #159, opened 2017-09-13
enable dot notation

mathisonian:
Enables use of syntax like `[Component.Subcomponent /]`


bclinkinbeard:
Looks like I forgot to update the .gitignore file when I ported everything to use src/dist. The dist folder shouldn't be added to Git in any of the packages.

mathisonian:
removed the `dist` folder

Issue #157, opened 2017-09-13
Performance concerns in scroll monitoring


mathisonian:
[These `getBoundingClientRect()` calls](https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/src/index.js#L39-L41) will cause reflows which may lead to some jank on scrolls. While we're making changes to the refs and scroll logic, it'd probably be worthwhile to see what can be done to avoid these calls. As I see it there are two options:

1) cache their values and update them on relevent events
2) lean more heavily on `scrollmonitor` and just use the values that it already calculates more directly

I think I'd be okay with (2), the only obvious downside being that it only handles vertical and not horizontal scrolling.

bclinkinbeard:
Dang, I didn't realize that caused a reflow. Lines 40 and 41 should happen at the top of `getRefs()` but I'm not sure how to get around the one on line 39. There is no event for a node having changed size, is there?

PR #160, opened 2017-09-13
Add percentInView to ref object

bclinkinbeard:
Also moves the one `getBoundingClientRect()` call out of the loop as mentioned in #157, though it still runs on every scroll event. :/

### DIFF
--- a/packages/idyll-document/src/index.js
+++ b/packages/idyll-document/src/index.js
@@ -29,6 +29,8 @@ const getScrollableContainer = el => {
 
 const getRefs = () => {
   const refs = {};
+  const containerNode = scrollContainer.item;
+  const containerRect = containerNode.getBoundingClientRect();
 
   scrollWatchers.forEach(watcher => {
     // get boolean props
@@ -40,13 +42,19 @@ const getRefs = () => {
 
     const domNode = watcher.watchItem;
     const rect = domNode.getBoundingClientRect();
-    const containerNode = scrollContainer.item;
-    const containerRect = containerNode.getBoundingClientRect();
+    const percentInView = Math.max(
+      0,
+      Math.min(
+        1,
+        (watcher.bottom - scrollContainer.viewportTop) / rect.height
+      )
+    )
 
     // left and right props assume no horizontal scrolling
     refs[domNode.dataset.ref] = {
       ...bools,
       domNode,
+      percentInView: Math.round(percentInView * 100),
       size: {
         width: rect.width,
         height: rect.height,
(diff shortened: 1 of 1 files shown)


packages/idyll-document/src/index.js
```diff
@@ -40,13 +42,19 @@ const getRefs = () => {
...
+        1,
+        (watcher.bottom - scrollContainer.viewportTop) / rect.height
+      )
+    )
 
     // left and right props assume no horizontal scrolling
     refs[domNode.dataset.ref] = {
       ...bools,
       domNode,
+      percentInView: Math.round(percentInView * 100),
```
mathisonian: Can we just pass the raw value for this, from 0 to 1?

bclinkinbeard: We can but the name would have to change. Open to suggestions.

mathisonian: `proportionInView`, `proportionVisible`, `fractionInView`, `visibility`. interestingly there is a whole [SE question dedicated to this](https://english.stackexchange.com/questions/275734/a-word-for-a-value-between-0-and-1-inclusive), although it isn't super helpful.

bclinkinbeard: lol nice, at least we're not alone. `fractionInView` isn't bad but when I see fraction I think numerator and denominator.

rreusser: First thought was `fractionInView`.

BBischof: As a slight modification on what was already suggestion: `portionInView`. It's a rational that is always a portion. I don't see a better word than `portion` to refer to it's containment in the unit-interval, and I think `rational` is redundant because you're not in a continuous context.

mathisonian: i like `portionInView`!

bclinkinbeard: Works for me. Commit incoming...

evanfrawley:
How is `portionInView` to be used on a component level? I'm having a bit of trouble understanding how it works

mathisonian:
Hi @evanfrawley, we ended up removing this API for performance reasons, but you can find more info on the reference object here: https://idyll-lang.org/docs/components/scrolling-and-refs.

The idea is that you can tag a component as a "reference", and Idyll will track certain information about it, making this info available to you in expressions. So for example, if I wanted to have some component play a video only when a some other component is above the viewport, I can do the following:

```
[ComponentA ref:"compA" /]

[ComponentB isPlaying:`refs.compA.isAboveViewport` /]
```

There's not a great way to to track the `portionInView` in a performant way at a system level, so that's something you'll have to implement in your JavaScript if you need it.

PR #161, opened 2017-09-13
clear require cache of components before SSR

### LINKED ISSUE #82: Changes to components aren't reflected in SSR during watch mode
If you change a component, the server doesn't automatically pull in the new code for SSR, even though it should. This can lead to things being slightly out of sync if you are changing custom component code

mathisonian:
this fixes #82


bclinkinbeard:
Sweet. I'm not familiar with `require.cache` or `require.resolve` but if it fixes that issue 👍 .

Issue #123, opened 2017-08-12
Bare derived and dataset passing


mathisonian:
One thing that seems to confuse people (understandably!) is that this works:

```
[var name:"x" value:10 /]
[component prop:x /]
```

but this doesn't.
```
[data name:"x" source:"file.json" /]
[component prop:x /]
```

this behavior was intentional to distinguish between passing something by value: ``[component prop:`x` /]``, and creating a 2-way binding to the variable ``[component prop:x /]``.

It seems like this is more confusing than it's worth, and the doc should accept datasets and derived variables without backticks, and figure out the right thing to do. We can show warnings if they try to update something that isn't a variable.

mathisonian:
Closed by https://github.com/idyll-lang/idyll/pull/163

PR #163, opened 2017-09-13
allow bare derived variables and datasets

mathisonian:
This updates the syntax to be more flexible, fixing an issue that was tripping a lot of people up (see  https://github.com/idyll-lang/idyll/issues/123)

It allows for properties to be passed like 

`[Component prop1:myVar prop2:myDerivedVar prop3:myDataset /]` instead of forcing

``[Component prop1:myVar prop2:`myDerivedVar` prop3:`myDataset` /]``

### DIFF
--- a/packages/idyll-document/src/utils/index.js
+++ b/packages/idyll-document/src/utils/index.js
@@ -95,17 +95,25 @@ const getVars = (arr, context) => {
     const [, [, nameValue]] = nameArr
     const [, [valueType, valueValue]] = valueArr;
 
-    if (valueType === 'value') acc[nameValue] = valueValue;
-    if (valueType === 'variable') acc[nameValue] = context[valueValue];
-    if (valueType === 'expression') {
-      const expr = valueValue;
-
-      acc[nameValue] = {
-        value: evalExpression(context, expr),
-        update: (newState, oldState) => {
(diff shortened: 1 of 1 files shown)


packages/idyll-document/src/utils/index.js
```diff
@@ -95,17 +95,25 @@ const getVars = (arr, context) => {
...
+          acc[nameValue] = evalExpression(context, expr);
+        }
+        break;
+      case 'expression':
+        const expr = valueValue;
+        acc[nameValue] = {
+          value: evalExpression(context, expr),
+          update: (newState, oldState) => {
+            return evalExpression(Object.assign({}, oldState, newState), expr)
+          }
```
bclinkinbeard: This could be a one line change couldn't it? Either way is fine though.

`if (valueType === 'variable') acc[nameValue] = context[valueValue] || evalExpression(context, expr);`

mathisonian: Yeah, I made that change, just because I thought there'd be more logic to add in. I can simplify it back (although I'm a little wary of that type of conditional assignment when 0 is a valid value for first expression)

bclinkinbeard: Ah, good point about 0. As you were... :)

PR #165, opened 2017-09-14
Strings have to be wrapped in quotes

bclinkinbeard:
Else things go boom when the expressions are evaluated


mathisonian:
Before we were doing `JSON.stringify()` for all the values (which takes care of wrapping strings in quotes). Do we need to add logic here for `Obects` as well, or is that handled elsewhere?   https://github.com/idyll-lang/idyll-interactive-document/blob/master/src/visitors/node.js#L93

bclinkinbeard:
Ah OK, I remembering wondering about that. Makes sense now. :) Will update.

bclinkinbeard:
There we go.

mathisonian:
👍 LGTM

PR #144, opened 2017-09-07
React 16

mathisonian:
This updates the project to use react-16 (targeted against the `idyll-doc-updates` branch). Its a relatively minor change, but will be good to get in, and will potentially allow us to take advantage of the new error boundary features.

There are some warnings being thrown by Victory (only in dev - not production) but they are aware and say they'll fix these in the next release https://github.com/FormidableLabs/victory/issues/724

PR #167, opened 2017-09-15
v2 prerelease prep

bclinkinbeard:
This basically combines #144 and #145 and does some dependency clean up and upgrades. I'm going to create the prerelease from this branch but also wanted a place to discuss a few questions. Will be making specific code/commit comments for those.

### DIFF
--- a/packages/idyll-document/src/index.js
+++ b/packages/idyll-document/src/index.js
@@ -100,11 +100,11 @@ class Wrapper extends React.PureComponent {
       updateRefsCallbacks.push(this.onUpdateRefs);
     }
 
-    this.state = { hasError: false, errorMessage: '' };
+    this.state = { hasError: false, error: null };
   }
 
   componentDidCatch(error, info) {
-    this.setState({ hasError: true, errorMessage: error.message });
+    this.setState({ hasError: true, error: error });
   }
 
   onUpdateProps(newState, changedKeys) {
@@ -177,7 +177,7 @@ class Wrapper extends React.PureComponent {
     if (this.state.hasError) {
       return (
         <div style={{ border: 'solid red 1px', padding: 10}}>
-          {this.state.errorMessage}
+          {this.state.error.message}
         </div>
       );
     }
--- a/packages/idyll-document/src/index.js
+++ b/packages/idyll-document/src/index.js
@@ -99,6 +99,12 @@ class Wrapper extends React.PureComponent {
     if (props.hasHook || exps.some(v => v.includes('refs.'))) {
       updateRefsCallbacks.push(this.onUpdateRefs);
     }
+
+    this.state = { hasError: false, errorMessage: '' };
+  }
+
+  componentDidCatch(error, info) {
+    this.setState({ hasError: true, errorMessage: error.message });
   }
 
   onUpdateProps(newState, changedKeys) {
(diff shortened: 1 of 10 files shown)


packages/idyll-document/src/index.js
```diff
@@ -99,6 +99,12 @@ class Wrapper extends React.PureComponent {
     if (props.hasHook || exps.some(v => v.includes('refs.'))) {
       updateRefsCallbacks.push(this.onUpdateRefs);
     }
+
+    this.state = { hasError: false, errorMessage: '' };
+  }
+
+  componentDidCatch(error, info) {
+    this.setState({ hasError: true, errorMessage: error.message });
```
bclinkinbeard: Should we store the whole error object instead? Also, and more importantly, I was getting a message about cross origin errors that you didn't seem to be. I'm not sure if this is due to a new React build or what. Any ideas?

packages/idyll-document/src/index.js
```diff
@@ -168,6 +174,14 @@ class Wrapper extends React.PureComponent {
   }
 
   render() {
+    if (this.state.hasError) {
+      return (
+        <div style={{ border: 'solid red 1px', padding: 10}}>
+          {this.state.errorMessage}
+        </div>
+      );
+    }
```
bclinkinbeard: I got rid of the `<pre>` because it was preventing wrapping.

packages/idyll-document/src/utils/index.js
```diff
@@ -61,30 +61,13 @@ const evalExpression = (acc, expr) => {
...
-        })()
-      };
+      return ${expr};
     })()
-  `);
+  `;
+
+  try {
+    return eval(e);
+  } catch (err) {}
```
bclinkinbeard: I split out the construction from the eval call for easier debugging. There have been various times when I needed to check what was actually being evaluated and this format makes that easier.

More importantly, removing the interior try catch means errors still make it to the console. I think our error bound will keep things from getting into a bad state but do we care about the console showing errors? Probably not but I wanted to be sure.

packages/idyll-document/src/index.js
mathisonian: Storing the whole object seems like a reasonable thing to do. This is the only thing I could find https://facebook.github.io/react/docs/cross-origin-errors.html#webpack, I guess I was running a production build and didn't see this, hmmm

packages/idyll-document/src/utils/index.js
mathisonian: I think thats okay - it is frustrating if a program is broken and it doesn't show any error messages

packages/idyll-document/src/index.js
bclinkinbeard: Agreed on the whole object.

packages/idyll-document/src/utils/index.js
bclinkinbeard: 👍

bclinkinbeard:
@mathisonian any objections to merging this? Turns out our build/publish setup is wholly inadequate, but I'll tackle that in a separate issue/PR.

mathisonian:
Nope, cool with me. I've also given you publish access to the compiler

bclinkinbeard:
👍 👍

PR #169, opened 2017-09-16
Build system to properly support ES workflows

### LINKED ISSUE #168: Create an adequate build system
Turns out supporting the `module` field in `package.json` is a little more involved than I realized. It should point to a file that is a module entry, but all the files it imports/re-exports should already be transpiled since tools like Webpack will not process them, they have/need their own Babel config, etc.

`react-router` has a pretty good setup that I think we can/should probably mimic. https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/tools/build.js and https://unpkg.com/react-router-dom@4.1.1/es/index.js for reference.

bclinkinbeard:
Fixes #168 

There were two things needed to enable consuming projects to actually take advantage of the `main` and `module` fields defined in the `package.json` files.

First, the source needed to actually be defined as ES modules, which it previously wasn't. This PR converts `idyll-components` and `idyll-document`. Not sure if we'll want/need to convert the others in the future.

Second, we needed to actually create an "ES bundle" via the build. This is essentially a copy of the ES source with everything but the module definitions/entry file transpiled ahead of time. This allows tree shaking in consumer builds without having to replicate our Babel setup.

Speaking of, the Babel setup for the `idyll-components` and `idyll-document` packages was streamlined and brought up to date. The `stage-2` preset was swapped out for the `babel-plugin-transform-object-rest-spread` plugin since I think that's all we were using from `stage-2`. We also need to use `BABEL_ENV` to create the two types of builds, so the `.babelrc` files now point to `.babelrc.js` files. This prepares us well for Babel 7, which will [deprecate](https://github.com/babel/babel/issues/5276) the use of `env` in `.babelrc` files.

This PR was modeled after `react-router`'s [setup](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/tools/babel-preset.js).

### DIFF
--- a/packages/idyll-compiler/package.json
+++ b/packages/idyll-compiler/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "nearleyc src/grammar.ne -o src/grammar.js",
     "test": "npm run build && mocha",
-    "prepublish": "npm run build"
+    "prepublishOnly": "npm run build"
   },
   "repository": {
     "type": "git",
--- a/packages/idyll-components/.babelrc
+++ b/packages/idyll-components/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["env", "stage-2", "react"]
+  "presets": [ "./.babelrc.js" ]
 }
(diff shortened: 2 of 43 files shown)


bclinkinbeard:
Thoughts on this @mathisonian? This one might be easiest to review one commit at a time.

bclinkinbeard:
Renegade merge incoming!

mathisonian:
👍 Thanks @bclinkinbeard - this all looks good and makes sense to me. I might ask you some general questions about using the module field in gitter

PR #177, opened 2017-09-25
Holy crap the tests are passing

bclinkinbeard:
Neither of these changes in isolation seemed to fix things, but tests are passing locally with these changes.

Jest back to v20.0.4 and `babel-jest` removed. (It's loaded automatically AFAIK.)


mathisonian:
🎉

Issue #179, opened 2017-09-26
Component property expressions not working as expected


mathisonian:
If I have include the following markup 

```
[Table data:`[{a: 'b'}, {a: 'c'}]` columns:`['a']` /]
```

The properties are passed into the Table component as strings rather than JavaScript expressions. The properties that the component sees look like this:

```
{ __expr__: { data: '[{a: \'b\'}, {a: \'c\'}]', columns: '[\'a\']' },
  data: '[{a: \'b\'}, {a: \'c\'}]',
  columns: '[\'a\']',
  children: [] }
```

mathisonian:
As far as I can tell, the problem is this - 

things work properly once there is an update because the `onUpdateProps` method gets called and triggers a state update on the wrapper component. However, the initial state doesn't include the evaluated info - as can be seen here: https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/src/index.js#L101

We'll need to make sure that the initial rendering has these expressions parsed properly.

mathisonian:
Patch in #182

PR #181, opened 2017-09-27
enzyme 3, react 16

mathisonian:
Work in progress - TODOs:

- [x] make sure the production react build imports in `idyll-cli` still work (i think they changed the path https://facebook.github.io/react/blog/2017/09/26/react-v16.0.html#packaging)
- [x] The `idyll-component` tests are passing. 
- [x] `idyll-document` tests are passing, but I was run into some issues with the `scrollcontainer` object when the component initializes - @bclinkinbeard you might have some insight. 
- [x] I've swapped out `react-micro-bar-chart` in the `idyll-cli` tests and replaced it with another component. However -- https://github.com/idyll-lang/idyll/issues/179 is still causing these tests to fail


mathisonian:
I think this should start passing all the tests once #182 is merged

bclinkinbeard:
Want to just pull #182 int? It's not very big. :)

mathisonian:
Pulled #182 in ✅

PR #182, opened 2017-09-28
Bugfix for expression initialization

mathisonian:
This fixes https://github.com/idyll-lang/idyll/issues/179 by modifying the conditional to include expressions like `` [Component prop:`{"static": "object"}` /]`` because otherwise these were not being properly initialized.

### DIFF
--- a/packages/idyll-document/src/index.js
+++ b/packages/idyll-document/src/index.js
@@ -98,19 +98,7 @@ class Wrapper extends React.PureComponent {
       updateRefsCallbacks.push(this.onUpdateRefs);
     }
 
-    const state = { hasError: false, error: null };
-    // const { __vars__, __expr__ } = props;
-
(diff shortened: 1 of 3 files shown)


bclinkinbeard:
@mathisonian can you explain why this works? I don't doubt it does, but it seems we're going from checking node prop values to checking their keys, and I don't quite understand why.

Also, we should probably squash this into one commit.

mathisonian:
By just looking at the keys It takes advantage of the fact that processing has already been done to understand which properties are attached to expressions or variables -- if a component has no expressions or variables, we don't need to wrap it.

Note this is slightly suboptimal because it adds wrappers for components that only have static expressions, but that doesn't seem too concerning

bclinkinbeard:
Oh nice, good thinking!

PR #186, opened 2017-09-30
use idyll-document module

mathisonian:
from @bclinkinbeard on [gitter](https://gitter.im/idyll-lang/Lobby):

> two weird things with the editor. it’s using the CJS builds of idyll-components and the editor’s start script is somehow watching and reacting to changes in idyll-document

1. I think this is "as expected" because we're bundling with browserify and it uses `main` or `browser` fields in package.json to resolve an entry point

2. Because we were importing idyll document directly from its source folder and not from the node_modules the changes were getting picked up. This updates the import to not point directly to the source

### DIFF
--- a/packages/idyll-editor/src/renderer.js
+++ b/packages/idyll-editor/src/renderer.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import * as components from 'idyll-components';
-import IdyllDocument from '../../idyll-document/src';
+import IdyllDocument from 'idyll-document';
 
 class Renderer extends React.PureComponent {
   render() {


bclinkinbeard:
Ah, good thinking. Lerna didn't like the non-relative import before `idyll-document` had been published. I'm going to work on a pull request that will enable us to work on the components and document and see the updates in the editor right away. It will be handy for development and better reflect real workflows. Part of that will likely be replacing `budo` with `webpack-dev-server`...

mathisonian:
Cool, that sounds good to me 👍

Issue #134, opened 2017-09-01
Support components within headings


mathisonian:
<a href="https://github.com/bclinkinbeard"><img src="https://avatars2.githubusercontent.com/u/140343?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [bclinkinbeard](https://github.com/bclinkinbeard)**
_Monday Jun 12, 2017 at 16:54 GMT_
_Originally opened as https://github.com/idyll-lang/idyll-compiler/issues/29_

----

`# Z: [DisplayVar var:`z` /].` does not actually parse the `DisplayVar` component and instead displays the tag as part of the header. I'd suspect this isn't unique to headings.

mathisonian:
Closing in favor of https://github.com/idyll-lang/idyll/issues/131

Issue #131, opened 2017-09-01
Parse headings recursively


mathisonian:
<a href="https://github.com/rreusser"><img src="https://avatars0.githubusercontent.com/u/572717?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [rreusser](https://github.com/rreusser)**
_Monday May 29, 2017 at 23:43 GMT_
_Originally opened as https://github.com/idyll-lang/idyll-compiler/issues/25_

----

This code:

```
## Test **test** *test*
```

Produces this output:

<img width="281" alt="screen shot 2017-05-29 at 19 42 57" src="https://cloud.githubusercontent.com/assets/572717/26563862/1fea3726-44a7-11e7-882e-431334b407f3.png">

Same as for `ul li` and similar, probably just need to parse the content string instead of just using it as-is.

PR #187, opened 2017-09-30
recursive parsing of headings

mathisonian:
this addresses https://github.com/idyll-lang/idyll/issues/131 (and #134) by updating "headers" in the AST to be recursively parsed. this means that bold and italic items are supported within headers now, along with components.


bclinkinbeard:
👏

PR #189, opened 2017-10-01
Convert editor build and watch modes to Webpack

bclinkinbeard:
Also adds `yarn run dev` to document and components so the editor will live reload changes in either of those packages.


mathisonian:
I'm getting some errors when trying to run this locally:

First, `extract-text-webpack-plugin` is not found and the dev-server doesn't launch. If I `yarn add` that, the dev-server launches, but I get this error in the browser:

```
Uncaught Error: Module build failed: Error: Unsupported preset format: undefined. (While processing preset: "/Users/mathisonian/projects/idyll/idyll/packages/idyll-components/.babelrc.js")
```

mathisonian:
Nevermind - some of that may have been due to changes I had locally

Issue #170, opened 2017-09-21
Define and implement "errorHelper" pattern


bclinkinbeard:
As discussed in #145 

> What if we checked for a prop on the wrapped component called errorHelper or something and, if it was defined, displayed it below the error? Component authors (especially us) could then define something to be displayed when the component has an error. It could list and describe supported props, show example syntax, etc. I think that could be super helpful.

bclinkinbeard:
In the course of experimenting with this it dawned on me that `IdyllDocument` doesn't actually need to be involved. I was thinking through how/if the error could be passed into the `errorHelper`, etc. and I realized the components can define their own error boundary. The only extra work this requires of the component dev is implementing a (likely one line, cookie cutter) `componentDidCatch` method. The bulk of the work is the error/help display, and that has to be done whether `IdyllDocument` is involved or not.

We can retain (and improve) the boundary already defined in `IdyllDocument`/`Wrapper` and it will act as a fallback for components that don't define their own.

Lastly, I love the idea of on-demand help. I confirmed this can easily be done already by checking `this.props.help` in the component. Currently, however, that has to be defined with a truthy value like `[Chart type:"scatter" help:true /]` or `[Chart type:"scatter" help:1 /]`. I think it would be ideal to support boolean props like `[Chart type:"scatter" help /]` but I don't think the compiler currently allows it.

Thoughts @mathisonian?

mathisonian:
Sounds good i think... could you post a more concrete example of what implementing it would look like in the component?

Having bare boolean props makes sense - i think we can work on that in a separate issue in the compiler. Let's start assuming that its done with `[Chart type:"scatter" help:true /]`, and then can concurrently work on adding the new syntax

bclinkinbeard:
Sure, any component that opts in would have to add something like this:

```
componentDidCatch(error, info) {
  this.setState({ hasError: true, error: error });
}

render() {
  if (this.state.hasError) return this.renderError();
  if (this.props.help) return this.renderHelp();
  // normal render code
}
```

Sounds good re: syntax

bclinkinbeard:
Put this in the "no time soon but someday" box. What if the links in the output, or some other UI controls, could be used to "autocomplete" the tag in the editor?

![image](https://user-images.githubusercontent.com/140343/31042350-c777ddcc-a573-11e7-8dc2-9a64cca92e0c.png)

mathisonian:
Fixed in #410

PR #190, opened 2017-10-01
Component with error bounds example

bclinkinbeard:
Example of how we could enhance the components to implement their own error bounds. The contents of both helper views are just POC but the overall structure should be replicable.

Per #170


bclinkinbeard:
Closing in favor of #201

PR #191, opened 2017-10-01
Improved IdyllDocument tests

bclinkinbeard:
This ended up being a mixed bag, not just tests :/

Edit: to clarify, this includes two "features".

- https://github.com/idyll-lang/idyll/pull/191/commits/748054ce0422d13de3ada78c572885fb8d9b3aeb ensures all custom components (as detected by a capitalized name) are wrapped. Previously, custom components that did not reference state, like `[Chart type:"scatter" /]`, were not wrapped. In order to provide an error boundary we have to wrap them.
- https://github.com/idyll-lang/idyll/pull/191/commits/eb31eff38304a094c928b06bb4832f86d6727acc adds support for a `src` prop to `IdyllDocument`. If an `ast` prop is not provided and a `src` prop is, `IdyllDocument` will compile the source on its own.

### DIFF
--- a/packages/idyll-document/src/index.js
+++ b/packages/idyll-document/src/index.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import scrollMonitor from 'scrollmonitor';
+import compile from 'idyll-compiler';
 import ReactJsonSchema from './utils/schema2element';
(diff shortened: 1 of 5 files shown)


mathisonian:
Cool, I like the `src` prop ✨

Issue #71, opened 2017-05-28
Expand test suite


mathisonian:
It's about time that we added a much more expansive test suite, given all the different ways in which Idyll may be used, and all the different options available. I think the first step for this is just to figure out what the structure looks like for having multiple projects in the tests folder (I believe @bclinkinbeard set this up so it should be fairly straightforward), then start testing different situations. 

Specifically, one thing I think we really need are different tests for when Idyll is included in the local node_modules, vs when it is used from the command line vs when it is used programatically. 

- `idyll`
  - [ ] command line options work
     - [ ] themes
     - [ ] component paths
     - [ ] minification and SSR
     - [ ] accepts strings or filepaths
     - [ ] watch mode
  - [ ] components
     - [ ] it finds custom components
     - [ ] it finds default components
     - [ ] it works with components from npm
     - [ ] it works with dot notation
  - [ ] server side rendering
     - [ ] variables render correctly
     - [ ] datasets render correctly
     - [ ] derived variables render correctly
- `idyll-components`
   - [ ] all components mount properly
   - [ ] all components mount properly outside of the browser (SSR)
   - [ ] what component specific tests??
- `idyll-document` - 
   - [ ] expressions work on initial mount and after updates
       - [ ] objects
       - [ ] strings 
       - [ ] variables
       - [ ] derived
       - [ ] datasets
   - [ ] basic AST renders properly
   - [ ] variable updating works
- `compiler` - tests are already pretty good there

mathisonian:
Things to cover

- `idyll`
  - [ ] command line options work
     - [ ] themes
     - [ ] component paths
     - [ ] minification and SSR
     - [ ] accepts strings or filepaths
     - [ ] watch mode
  - [ ] components
     - [ ] it finds custom components
     - [ ] it finds default components
     - [ ] it works with components from npm
     - [ ] it works with dot notation
  - [ ] server side rendering
     - [ ] variables render correctly
     - [ ] datasets render correctly
     - [ ] derived variables render correctly
- `idyll-components`
   - [ ] all components mount properly
   - [ ] all components mount properly outside of the browser (SSR?)
   - [ ] what component specific tests??
- `idyll-document` - 
   - [ ] expressions work on initial mount and after updates
       - [ ] objects
       - [ ] strings 
       - [ ] variables
       - [ ] derived
       - [ ] datasets
   - [ ] variable updating works
- `compiler` - tests are already pretty good there

mathisonian:
[...]

PR #193, opened 2017-10-03
More tests

mathisonian:
This is a PR to close out the list in https://github.com/idyll-lang/idyll/issues/71. So far its just additions to the `idyll-document` tests

 
- `idyll`
  - [x] options work
  - [x] components
  - [x] server side rendering
- `idyll-components`
   - [x] all components mount properly
   - [x] all components mount properly outside of the browser (SSR)
- `idyll-document` 
  - [x] expressions work on initial mount and after updates
  - [x] basic AST renders properly
  - [x] variable updating works


packages/idyll-document/test/vars.js
```diff
@@ -2,26 +2,34 @@ import React from 'react';
...
 let doc;
+let component;
+
+const FAKE_DATA = 'FAKE DATA';
 
-beforeEach(() => {
-  doc = new InteractiveDocument({ast, components, datasets: {myData: 'FAKE DATA'}});
+beforeAll(() => {
+  // doc = new InteractiveDocument({ast, components, datasets: {myData: FAKE_DATA}});
+  component = mount(<InteractiveDocument ast={ast} components={components} datasets={{myData: FAKE_DATA}} />);
```
bclinkinbeard: Rename to IdyllDocument? :)

mathisonian:
@bclinkinbeard do you have scripts or anything that you were using for the the fixtures in `idyll-document`? I'm hoping to update the `src.idl` file with a few things and have the schema and AST reflect that.

bclinkinbeard:
@mathisonian there is https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/src/utils/idl2ast.js. I had actually started on making that a bit more generic and reusable but got sidetracked.

packages/idyll-document/package.json
```diff
@@ -15,7 +15,7 @@
     "build": "npm run build:cjs && npm run build:es",
     "dev": "yarn run build:es --watch",
     "prepublishOnly": "npm run build",
-    "test": "jest"
+    "test": "jest --updateSnapshot"
```
bclinkinbeard: We don't want to pass `--updateSnapshot` every time, it will prevent tests from ever failing. That should either be a manual step or an separate script.

mathisonian: Oh yeah thanks - I forgot to remove that!

bclinkinbeard:
I hate the code review UX. I will do a real review of this PR once it's ready.

mathisonian:
Cool - I think with the latest push most of https://github.com/idyll-lang/idyll/issues/71 is addressed. A couple things I'd like to add but maybe not for this PR necessarily - 

* Updating the `src.idl` in idyll-document is a little brittle, and requires some manual adjustments. It would be great to get to a spot where we automate that and rely more on enzyme selectors to target specific nodes in the document.

* We still don't test the CLI directly, rather we test the API that it uses. I'm not sure if theres a great way to do this that doesn't involve spawning processes, but it feels a little lacking currently.

mathisonian:
@bclinkinbeard the last commit upgrades victory and sets the props so it has deterministic ids.

packages/idyll-components/test/components.js
```diff
@@ -4,82 +4,256 @@ import Adapter from 'enzyme-adapter-react-16';
...
+      it('<Container />', () => {
+        const wrapper = mount(<Container />);
+      });
+      it('<Preload />', () => {
+        const wrapper = mount(<Preload />);
+      });
+      it('<Screen />', () => {
+        const wrapper = mount(<Screen />);
+      });
+    });
```
bclinkinbeard: Nothing is being asserted in this file... If we're just making sure they don't blow up maybe we should use `expect(shallow(<Screen />)).toNotThrow()` or whatever?

packages/idyll-document/test/vars.js
```diff
@@ -30,11 +36,86 @@ describe('Component state initialization', () => {
...
+      id: 'dynamicObjectDisplay',
+      html: `<span>${JSON.stringify({dynamic: 2.0})}</span>`
+    }, {
+      id: 'dataDisplay',
+      html: `<span>${FAKE_DATA}</span>`
+    }];
+
+    checks.forEach((check) => {
+      const display = displayComponents.find(`#${check.id}`);
+      expect(display.length).toBeGreaterThan(0);
```
bclinkinbeard: Nit: `toBe(1)`?

bclinkinbeard:
Agreed. I want to make it much easier to test .idl source variations. I think part of that will include leaning more heavily on snapshots instead of manually maintaining fixtures. Definitely a separate PR though.

PR #197, opened 2017-10-04
Move Enzyme config into setup files

bclinkinbeard:
I discovered this is possible. Hides the ugly config in a single setup file.


mathisonian:
Cool ✨

PR #199, opened 2017-10-05
remove arrow function from compiler

mathisonian:
cc @bclinkinbeard 

Can you comment on whether this was blowing up because (1) uglify doesn't like arrow functions and the rest of the code that uses them works because it is transpiled or (2) the nearly library doesn't like arrow funcs for some reason

### DIFF
--- a/packages/idyll-compiler/.babelrc
+++ b/packages/idyll-compiler/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["./.babelrc.js"]
+}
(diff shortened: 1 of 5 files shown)


bclinkinbeard:
UglifyJS flat out doesn't support ES6+ so I'm guessing that's the only issue

mathisonian:
Okay i've updated to add a babel build to the compiler so hopefully that takes care of the problem

bclinkinbeard:
Now I'm getting `Invalid assignment [/Users/bclinkinbeard/Code/idyll/packages/idyll-compiler/src/grammar.js:169,0]`, which is weird since it shouldn't be using that file anymore. I'm running this branch, the dist files are built... not sure what's going on.

mathisonian:
I needed to move the `index.js` file into src for it to get babel'd too. Weird that it loaded the module at all without that

bclinkinbeard:
🎉 Success! 🎉

PR #201, opened 2017-10-06
Example component with error boundary and help

bclinkinbeard:
@mathisonian thoughts on adopting this general pattern? Deciding how we want to do this and applying it to all the built-in components is one of the larger things left before a v2 release.

### DIFF
--- a/packages/idyll-components/src/chart.js
+++ b/packages/idyll-components/src/chart.js
@@ -18,9 +18,56 @@ class Chart extends React.PureComponent {
   constructor(props) {
     super(props);
     this.id = chartCount++;
+    this.state = {};
+    this.renderError = this.renderError.bind(this);
+    this.renderHelp = this.renderHelp.bind(this);
+  }
+
+  componentDidCatch(error, info) {
+    this.setState({ hasError: true, error: error });
+  }
+
+  renderError() {
+    return ([
+      <div style={{
+        padding: '1em',
+        fontSize: '1.5em',
+        fontWeight: '900',
+        color: 'darkred',
+        backgroundColor: 'pink'
+      }}>
+        {this.state.error.message}
+      </div>,
+      <br/>,
+      this.renderHelp()
+    ])
+  }
+
+  renderHelp() {
+    return (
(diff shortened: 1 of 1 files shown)


mathisonian:
I'm not convinced it makes sense to define the render error message individually for every component rather than including that in the wrapper. What's the advantage of doing it this way? Beyond less code, an advantage of doing it the wrapper is that custom components that don't implement an error handler will still have this baked in.

For the help message, I think the ideal for me would be if we could generate these based on `propTypes`, though I haven't thought through exactly how to implement it. That way in the default components (and user-defined custom components) all they would need to do is define the `propTypes` and then error handling and help messages as expected.

bclinkinbeard:
I guess you're probably right that we can centralize the error handling/display. It seems unlikely that a component would need or want to display custom content, 99.9% of the time you'd just want to show `error.message`. Also, if components implement their own `componentDidCatch` I think it will just override the default handling anyway.

Interesting idea with `propTypes` for the help! I'm going to play with that now.

bclinkinbeard:
https://github.com/idyll-lang/idyll/pull/201/commits/64716dab156c1a180acba1bf1679195cadfe3baf adds (POC) automated help based on `defaultProps`. Turns out `propTypes` doesn't give you much useful info at runtime, all you get are the keys. Downside is obviously the question what about props without default values?

bclinkinbeard:
Maybe we just need to define a set of prioritized fallbacks for each scenario where we use the first one that exists.

**Errors;** `Comp.errorUI() | Wrapper.errorUI()`
**Help:** `Comp.helpUI() | {...Comp.propTypes, ...Comp.defaultProps}`

mathisonian:
That model makes a lot of sense to me. Too bad PropTypes isn't so useful - seems like there should be some way to get the info out, at least when `NODE_ENV = debug`

bclinkinbeard:
I was surprised too but apparently it's a FOL. I will tinker with the model referenced above.

mathisonian:
Closing this as it was superseded by #410

PR #237, opened 2017-12-24
fix regexs for detecting keys that look like callbacks

joshuahhh:
`"_regionOfInterest".match(/on[A-Z].*/)` returns a match.
`"_regionOfInterest".match(/^on[A-Z].*/)` doesn't.
Hooray!


mathisonian:
Thanks @joshuahhh!

PR #252, opened 2018-02-12
FullWidth / TextContainer integration

mathisonian:
This adds a `fullWidth` prop to components. The compiler sees this flag and breaks them out of the standard text-container, making the CSS/styling much more straightforward. 

Example,

```
... standard boilerplate...


[CustomD3Component fullWidth:true state:state /]
[button onClick:`state++`]
  Click Me.
[/button]
```

produces:

screen shot 2018-02-12 at 12 42 40 am

This can also be used compositionally as a component:

```
Normal text container
...


[FullWidth]
   // my full width content
[/FullWidth]

Normal text container
...
```

and it exposes a `[TextContainer /]` component for authors to use if they want to manually reset some text in the normal place. For example, to create a custom component with text in the container as normal and some sort of full bleed background:

```
[FullWidth]
   [CustomComponent /] 
   [TextContainer]
     Normal text...
   [/TextContainer]
[/FullWidth]
```

Generally it makes things much cleaner, and particularly removes the need for hacks like https://github.com/idyll-lang/idyll/blob/master/packages/idyll-components/src/utils/container.js#L34-L48 that need to do size calculations on the client, and are hard to do correctly responsively.


mathisonian:
Tests and more cleanup to come.

/cc @bclinkinbeard @rreusser

mathisonian:
I think this is ready to merge ✨ 

- Tests are updated and passing. 
- Everything is backwards compatible, so it can just be a minor feature bump
- It continues some of the work that was started in https://github.com/idyll-lang/idyll/pull/247 towards making `themes` and `layouts` more modular and friendlier to work with.

In my testing this setup removes a lot of headache that we encountered with the first round the `Feature` component, and removes the need to mess around with calculating negative margins, etc in the CSS.

PR #256, opened 2018-03-04
Fix bug with passing props down to <button>

hmuar:
Fix small `ReferenceError: props is not defined` bug introduced by previous commit to Button component.

### DIFF
--- a/packages/idyll-components/src/button.js
+++ b/packages/idyll-components/src/button.js
@@ -2,8 +2,9 @@ import React from 'react';
 
 class Button extends React.PureComponent {
   render() {
+    const { onClick, updateProps, ...props } = this.props;
     return (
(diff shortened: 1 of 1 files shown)


hmuar:
Ah it's tripping up on the can't find 'react-latex' error. I was getting that a bit earlier locally when i created a new project, was going to look into it, but any subsequent new projects I create no longer hit that error. I guess something to do with moving to `mathisonian/react-latex`?

mathisonian:
Thanks @hmuar - I'm going to merge since this looks good and investigate why those tests are failing separately.

mathisonian:
I released a patch with this to npm - I'm tracking that test issue here: https://github.com/idyll-lang/idyll/pull/257

PR #266, opened 2018-03-10
Fix CSS content escaping in github theme

rreusser:
The github CSS theme had a couple `content` rules with improperly escaped characters. The result was stray characters. For example, this idyll

```markdown
This file is named `index.idl`
```

was rendered as:

screen shot 2018-03-10 at 11 35 27 am

This fix renders it as:

screen shot 2018-03-10 at 11 35 21 am

I also changed `.pl-c2::before` to the unicode character for `^M`, though I don't know what sort of markdown actually triggers this rule. I suspect it was supposed to be the special character sequence for a carriage return but accidentally got flattened into a caret followed by a capital M, which is something different.

### DIFF
--- a/packages/idyll-themes/src/github/styles.js
+++ b/packages/idyll-themes/src/github/styles.js
@@ -79,7 +79,7 @@ body {
 }
 
 .pl-c2::before {
-  content: "^M";
+  content: "\\000d";
 }
 
 .pl-sr .pl-cce {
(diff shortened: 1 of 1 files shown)


mathisonian:
Thanks @rreusser! I'll release a patch to npm with these updates.

These github styles are extracted from https://github.com/sindresorhus/github-markdown-css, which also seems to have that `^M`, I'm not totally sure what it is

mathisonian:
`v2.3.1` is published

Issue #269, opened 2018-03-11
Emphasis after list rendered incorrectly


rreusser:
This code:

```markdown
- foo
- bar

*Wow!*
```

is rendered as:

<img width="116" alt="screen shot 2018-03-11 at 12 51 10 pm" src="https://user-images.githubusercontent.com/572717/37257807-0c02850a-252b-11e8-91c1-4ccd3e557e3d.png">

A workaround is:

```markdown
- foo
- bar

[em]Wow![/em]
```

Which is rendered as:

<img width="97" alt="screen shot 2018-03-11 at 12 53 19 pm" src="https://user-images.githubusercontent.com/572717/37257816-31c0284c-252b-11e8-983b-61df78736c98.png">

It seems it's greedily interpreting the `*` as the start of a new list item.

PR #271, opened 2018-03-11
Require space after unordered or ordered list bullet

rreusser:
This PR modifies list parsing to *require* whitespace after the list bullet (`-`, `*`, or `1.`). This is a bit closer to github style markdown at least. This may be a partial solution of #269, though I'm not 100% certain this will get it. I think it might be a step in the right direction though.


mathisonian:
🤘 thanks @rreusser - version bumped on npm

PR #274, opened 2018-03-17
add initial state property

mathisonian:
This lets `IdyllDocument` users specify an initial state programatically. This is useful for cases like persisting document state across editor updates (or pulling from a stateful url, etc) where you want the document to initialize in a way that differs from the initial values specified in the idyll layout itself.

Example:

```
[var name:"x" value:1 /]
[derived name:"xSquared" value:`x * x` /]

[display var:x /],  [display var:xSquared /]
```
// output -> 1, 1

but if you initialized idyll's state programatically:

```js
<IdyllDocument initialState={{x: 2}} />
```

then the output will be `2, 4` on initial render.

### DIFF
--- a/packages/idyll-document/src/index.js
+++ b/packages/idyll-document/src/index.js
@@ -199,10 +199,10 @@ class IdyllDocument extends React.PureComponent {
     const Wrapper = createWrapper({ theme: props.theme, layout: props.layout });
 
 
-    const initialState = {
+    const initialState = Object.assign({}, {
       ...getVars(vars),
       ...getData(data, props.datasets),
-    };
+    }, props.initialState ? props.initialState : {});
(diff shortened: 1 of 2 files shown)


packages/idyll-document/src/index.js
```diff
@@ -199,10 +199,10 @@ class IdyllDocument extends React.PureComponent {
     const Wrapper = createWrapper({ theme: props.theme, layout: props.layout });
 
 
-    const initialState = {
+    const initialState = Object.assign({}, {
       ...getVars(vars),
       ...getData(data, props.datasets),
-    };
+    }, props.initialState ? props.initialState : {});
```
bclinkinbeard: `props.initialState || {}` FTW

Issue #70, opened 2017-05-27
Static analysis plugins


rreusser:
Prior art: [remark](https://github.com/wooorm/remark)

Goal: Table of contents. References. Other similar things.

Proposed interface for plugins that are applied to the AST after parsing and before use:

```
module.exports = {
  name: 'tableOfContents',
  plugin: function (ast) {
    return {headings: ...}
  }
};
```

Presumably it can do the cowboy thing and even modify the AST. Like to add anchors to the headings. The data could get registered globally as:

```
{pluginData: {tableOfContents: {headings: ...}}}
```

So that you could write:

```
[TableOfContents/]
```

Which would know enough to grab the `tableOfContents` data and render it to a list. This would also work great for references. You could write:

```
As seen in [Reference author:"Smith"/]
```

And then write at the end of the document

```
[References/]
```

And again…………… magic?

The other component to this would be a couple `visitor` style helper functions to make it easier to traverse the AST. Presumably to use these you'd write `idyll index.idl --plugin table-of-contents` which would have to get fancy about finding and requiring it via `node_modules`. Seems possible though.

rreusser:
Or just thinking out loud, you there could be a `visitor` interface where you're write:

```
visitors: {
  'h1': function sectionVisitor (node) {
    var name = node[2];
    var anchor = parameterize(name);
    node[1].push(['id', anchor])
 
    var subsections = traverse(node, subsectionVisitor);

    this.sections.push({name, anchor, subsections});

    // Don't recurse:
    return false;
  }
}
```

mathisonian:
I like this a lot, it seems like it could be pretty powerful. What I haven't quite figured out is where this code would live, and how the user would interact with it. Take the example of a table of contents plugin - the plugin code itself presumably exists as a module that users can install, does that module also include a component that users then include in the markup? 

Counterpoint - you could probably get a lot of this functionality by making smarter components, e.g. a `References` component that knows every time a `Cite` component is invoked

dundalek:
I like this project a lot and a support for plugins would be needed for my particular use case.

Will share some thoughts out loud too. I imagine the plugins would be able to hook into the build pipeline. There could be hooks for different stages like pre-parse, AST traversal, post-output, etc. User could include and configure plugins maybe via markdown front matter config.

[...]

PR #280, opened 2018-03-28
Compiler plugins

mathisonian:
This adds support for custom compile-time plugins. This feature has been discussed in #70 quite a bit, and should enable fun things like references, tables of contents, etc. Some other use cases could be shelling out to other languages like r/python (or DSLs like vega-lite, penrose) to generate charts at compile time, fetching dynamic data, code linting, spellchecking, etc.

I've added a new module, [idyll-ast](https://github.com/idyll-lang/idyll/tree/compiler-plugins/packages/idyll-ast), that will be useful for plugin authors who want to manipulate the AST somehow. This module is factored out of the compiler, which now depends on it, so it should be pretty stable.

Here's how it works: you define a plugin and the plugin receives the AST as input. It can modify this however it wants, and the results will be passed on to the runtime. A very simple example:

```js
const AST = require('idyll-ast');

module.exports = (ast) => {
  return AST.appendNode(ast, AST.createNode('div', {}, 'Hello World!') );
};
```

That is all. Right now it is expected that each of the plugins will be a standalone JS module. You can configure Idyll to use plugins in the following ways:

*via package.json*
```
"idyll": {
  "compiler": {
    "postProcessors": ['my-idyll-postprocessor']
  }
}
```

*via API*
```js
var Idyll = require('idyll');

var idyll = Idyll({
    compiler: {
       postProcessors: ['my-idyll-postprocessor']
    }
});

```

*in browser (manually)*
```js
const compile = require('idyll-compiler');
const myPostprocessor = require('my-idyll-postprocessor');
const IdyllDocument = require('idyll-document');

const ast = compile(idyllMarkup, { postProcessors: [ myPostprocessor ] });

return <IdyllDocument ast={ast} />
```

I'm happy to make API changes if people have preferences. I'll be testing this a bit more before merge. 

Notifying folks who participated on the previous discussion: 

/cc @rreusser @dundalek @heathermiller @bclinkinbeard


rreusser:
Main question is about whether async behavior is a natural extension since there are a number of use-cases. (fetching remote data, inlining base64 images, more?) It's fine if the answer is "no", but just curious.

rreusser:
(Oh, and also this is awesome! <3)

mathisonian:
Great point - we probably do want to support async. One option would be to determine if things are async would be to check if the plugin takes a callback, e.g. 

*async* 
```js
const AST = require('idyll-ast');

module.exports = (ast, cb) => {
  cb(AST.appendNode(ast, AST.createNode('div', {}, 'Hello World!') ));
};
```

or just check if the function provided is async? This feels like a little fancy but maybe people are used to the syntax by now?

*async* 
```js
const AST = require('idyll-ast');

module.exports = async (ast) => {
  const myValue = await fetch(...);
  return AST.appendNode(ast, AST.createNode('div', {}, myValue) ));
};
```

rreusser:
TBH I've never actually used `async`, but I have nothing in particular against it except for the fact that it's nothing you can't accomplish with callbacks. 😄

mathisonian:
The last commit adds support for asynchronous plugins via callbacks: 

```
module.exports = (ast, cb) => {
  cb(err, updatedAST);
};
```

The compiler is completely synchronous ATM, so this change required updating the API to support asynchronous usage - I'll update the dependents to reflect this.

mathisonian:
This is published in `idyll@2.7.0`, and I'll be adding a section to the docs site to match.

Issue #250, opened 2018-02-12
Command line build starts very slowly


mathisonian:
Seems to be due to babel trying to compile react?

```
[BABEL] Note: The code generator has deoptimised the styling of "/Users/mathisonian/projects/equation-blog/node_modules/react-dom/cjs/react-dom.development.js" as it exceeds the max of "500KB".
```

PR #287, opened 2018-04-06
Updated Babelify config to ignore node_modules.

andrewosh:
Hey Idyllers,

Adding this teeny bit of configuration to Babelify should get it to ignore any packages in all relevant `node_modules` directories. As of now, it's processing large files in the initial build (as described in #250). This does seem to speed up the first build quite a bit, but we should test this on a handful of projects to make sure Implicit Babelification of node_modules can be safely removed.

Also, I'll use this PR opportunity to introduce myself to @bclinkinbeard and @rreusser! I've quietly followed Idyll development since the beginning -- super cool to see how far it's come. Excited to contribute!

### DIFF
--- a/packages/idyll-cli/src/pipeline/bundle-js.js
+++ b/packages/idyll-cli/src/pipeline/bundle-js.js
@@ -28,12 +28,31 @@ const toStream = (k, o) => {
   return s;
 };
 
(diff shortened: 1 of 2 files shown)


mathisonian:
👍 LGTM, will merge once the tests pass

Issue #246, opened 2018-02-10
Component metadata


mathisonian:
It would be nice if components could register some meta-data when they get installed.

A concrete example:

The [idyll-apparatus-component](https://github.com/idyll-lang/idyll-apparatus-component) allows users to integrate Apparatus into their idyll docs. However, in order to use the component, after installing it they either have to write

```
[IdyllApparatusComponent /]
```

instead of just 

```
[Apparatus /]
```

or they have to explicitly configure that mapping themselves. It would be much nicer for the component to be able to expose an idyll name field that differed from the package name.

Another use case would be specifying properties of the component, for example see https://github.com/idyll-lang/idyll/issues/245

/cc @bclinkinbeard @joshuahhh

mathisonian:
The docs should also use this metadata to generate pages on idyll-lang.org.

Issue #118, opened 2017-08-09
Allow components to opt into (or out of SSR)


mathisonian:
It would be cool if Idyll was smart enough to know which components should be rendered during the SSR preprocessing step. For example, it may be the case that custom components don't play well with SSR (e.g. https://github.com/idyll-lang/idyll/issues/117), but having one component in your document that doesn't pre-render shouldn't cause the whole document to abort server side rendering.

I'm not sure what the best approach to this would be. Perhaps setting a flag on the component prototype?

mathisonian:
I wonder if upgrading to react 16 will provide an easy fix for this, since they've changed the way that errors in individual components affect the entire tree's rendering 

/cc @bclinkinbeard

mathisonian:
This could be specified via the mechanism described in https://github.com/idyll-lang/idyll/issues/246

PR #288, opened 2018-04-10
Component metadata

mathisonian:
This PR adds metadata to the standard library of components, so that Idyll authoring tools can provide better guidance, feedback, and error messages. It also fixes a handful of bugs and inaccuracies in the documentation that I found along the way.

### DIFF
--- a/packages/idyll-components/src/button.js
+++ b/packages/idyll-components/src/button.js
@@ -15,4 +15,14 @@ Button.defaultProps = {
   onClick: function() {}
 };
 
+Boolean._idyll = {
+  name: "Boolean",
+  tagType: "open",
+  children: ['Click Me.'],
+  props: [{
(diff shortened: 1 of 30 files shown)


mathisonian:
This provides a mechanism for adding functionality like https://github.com/idyll-lang/idyll/issues/246 and https://github.com/idyll-lang/idyll/issues/118.

packages/idyll-components/src/button.js
```diff
@@ -15,4 +15,14 @@ Button.defaultProps = {
   onClick: function() {}
 };
 
+Boolean._idyll = {
```
andrewosh: `Boolean` -> `Button`

mathisonian: Fixed & patched, thanks

Issue #146, opened 2017-09-08
Editor resets initial state


bclinkinbeard:
If you change the doc via interaction before focusing your cursor in the editor the output is rest

mathisonian:
This can be fixed using https://github.com/idyll-lang/idyll/pull/274

mathisonian:
Fixed by #289

PR #289, opened 2018-04-10
Enhance idyll-document logic

mathisonian:
This PR

* adds logic to `idyll-document` to persist state across markup updates
* it also adds logic to persist the previous output in the case of a compiler error
* fixes a bug in the editor on the docs site


mathisonian:
This fixes https://github.com/idyll-lang/idyll/issues/146

PR #290, opened 2018-04-10
Equation duplication

### LINKED ISSUE #138: Equations duplicatedduplicated in static html.
<a href="https://github.com/rreusser"></img></a> **Issue by [rreusser](https://github.com/rreusser)**
_Thursday Jun 08, 2017 at 20:48 GMT_
_Originally opened as https://github.com/idyll-lang/idyll-default-components/issues/15_

----

`[Equation]y = x[/Equation]` renders like `y=xy=x` in the static html because there's a html and a mathml version. It's a result of the katex css getting deferred to after the js has run and then after it's fetched the css. It might be nice if the css insert were optional so that you could inline it yourself into the page's css.

mathisonian:
This prevents the equation component from inserting the CSS tag if it has already been appended to the document. It also adds an option `skipCSS` to force the component to avoid inserting it.

It updates the HTML template that idyll uses to include the tex css file when necessary, to fix #138.

### DIFF
--- a/packages/idyll-cli/src/client/_index.html
+++ b/packages/idyll-cli/src/client/_index.html
@@ -36,6 +36,9 @@
     {{#twitterHandle}}
       <meta property="twitter:creator" content="{{twitterHandle}}">
     {{/twitterHandle}}
+    {{#usesTex}}
+      <link rel="stylesheet" href="styles.css">
+    {{/usesTex}}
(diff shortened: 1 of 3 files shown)


mathisonian:
This is available in the latest NPM release, @rreusser, let me know if you still have any issues here.

PR #292, opened 2018-04-11
Extract resolver behavior

andrewosh:
Moving the data/component/CSS resolution behavior into a self-contained set of small modules will make it easier to support extending these behaviors in the future. 

This PR includes the following changes (which can be broken up into a separate set of PRs if desired):
1. The introduction of the `ComponentResolver`, `DataResolver`, and `CSSResolver` classes.
2. The addition of the `debug` module (gated by an environment variable), which gives a bit more insight into which files are being included in the build process.
3. The addition of `src/errors.js`, which is the beginning of a consolidated set  of `Error` classes.
4. `paths.DATA_DIR` is now included in the set of watched directories -- a full rebuild will be triggered whenever data is is modified.
5.  Support for loading custom component and default components from multiple directories.

As of now `CSSResolver` and `DataResolver` are very minimal, and have just been ported into Resolvers to present a consistent interface.

`ComponentResolver` contains a slightly modified approach to component resolution, but the underlying algorithm is the same as before.

Any thoughts on the overall approach?


packages/idyll-cli/bin/idyll.js
```diff
@@ -1,6 +1,7 @@
 #! /usr/bin/env node
 
 const idyll = require('../src/');
+const debug = require('debug')('idyll-cli')
```
mathisonian: Maybe use `idyll:cli` convention if we want to re-use debug across other modules

packages/idyll-themes/package.json
```diff
@@ -41,5 +41,8 @@
     "cross-env": "^5.0.5",
     "jest": "^20.0.4",
     "rimraf": "^2.6.2"
+  },
+  "dependencies": {
+    "debug": "^3.1.0"
```
mathisonian: Don't add these unused deps

packages/idyll-cli/src/pipeline/parse.js
```diff
@@ -14,102 +20,25 @@ const getFilteredAST = (ast) => {
...
-    return acc;
-  }
-
-  return tree.reduce(
-    byName,
-    []
-  )
+exports.getComponentNodes = (ast) => {
+  const ignoreNames = ['var', 'data', 'meta', 'derived'];
+  return filterNodes(ast, s => !ignoreNames.includes(s));
```
mathisonian: I believe this will return the nodes still nested, when we really want to return a flattened list

packages/idyll-cli/src/pipeline/parse.js
```diff
@@ -126,8 +55,10 @@ exports.getDataJS = (ast, DATA_DIR, o) => {
       );
 
       if (source.endsWith('.csv')) {
+        debug(`Loading ${source} as a CSV into data variable ${name}`)
         acc[name] = Papa.parse(slash(path.join(DATA_DIR, source)), { header: true }).data;
       } else if (source.endsWith('.json')) {
+        debug(`Loading ${source} as a JSON document into data variable ${name}`)
         acc[name] = require(slash(path.join(DATA_DIR, source)));
       } else {
         throw new Error('Unknown data file type: ' + source);
```
mathisonian: Still using old style error

mathisonian:
Thanks @andrewosh, I think this approach makes sense - I like splitting out the resolvers like that and ✨ for getting multiple component paths working.

andrewosh:
The current resolver behavior (as of https://github.com/idyll-lang/idyll/pull/292/commits/c1ccf4263eefe73598d177574063326481485958) works for all existing tests (locally, still waiting on CI...), and for an additional test in `test/multiple-component-dirs/` that tests multiple custom components directories. 

As a small additional change, the `findNodes` function was added to `idyll-ast`, which returns a flattened list of nodes matching a selector function.

packages/idyll-cli/src/pipeline/parse.js
```diff
@@ -227,6 +127,7 @@ exports.getBaseHTML = (ast, template) => {
 
 exports.getHTML = (paths, ast, _components, datasets, template, opts) => {
   const components = {};
+  console.log('_components:', _components)
```
mathisonian: `debug()`

packages/idyll-cli/src/resolvers/components.js
```diff
@@ -0,0 +1,130 @@
...
+   *  1) The default components.
+   *  2) The custom components.
+   *  3) The user-defined aliased components.
+   *
+   * Each new components list potentially overwrites previously-resolved components, according to the
+   * above order of precedence.
+   */
+  _loadPaths() {
+    var componentsMap = new Map();
+    var prioritizedDirs = [this.paths.DEFAULT_COMPONENT_DIRS, this.paths.COMPONENT_DIRS];
```
mathisonian: any reason for preferring `var`s here?

packages/idyll-cli/src/resolvers/components.js
```diff
@@ -0,0 +1,130 @@
...
+  resolve(name) {
+    const self = this
+    const candidates = [pascalCase(name), paramCase(name), name.toLowerCase()];
+    debug(`Searching for component: ${name} with candidates: ${candidates}`);
+
+    var resolved = null
+
+    candidates.forEach(name => {
+      // Once one of the candidates has been found, don't continue searching.
+      if (resolved) return
```
mathisonian: This is maybe better since its explicit but `Array.some()` can also be useful for short circuiting iteration

packages/idyll-cli/src/resolvers/data.js
```diff
@@ -0,0 +1,50 @@
...
+
+  resolve(name, source) {
+    debug(`Resolving data with name ${name} and source ${source}`);
+
+    if (name[0] !== 'value') {
+      throw new errors.UnsupportedDataTypeError(name[0]);
+    }
+    if (source[0] !== 'value') {
+      throw new errors.UnsupportedDataTypeError(source[0]);
+    }
```
mathisonian: 👍

packages/idyll-cli/src/resolvers/components.js
andrewosh: Nope, just used to writing `var` -- changed for consistency

packages/idyll-cli/src/resolvers/components.js
```diff
@@ -0,0 +1,130 @@
...
+   *
+   * 1) If a component alias exists, then use the aliased path.
+   * 2) If a custom component exists, then use that component's path.
+   * 3) If a default component exists, use its path.
+   * 4) If no component has been found, attempt to resolve a component installed through NPM.
+   * 5) If the NPM resolution failed, check if this is a valid HTML tag, and use the tag name.
+   * 6) Else, return nothing (this is a failure).
+   */
+  resolve(name) {
+    const self = this
```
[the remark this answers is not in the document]
andrewosh: :+1: reflex

mathisonian:
LGTM once the tests pass.

The only thing that requires a documentation update is the ability to match multiple component directories, correct?

PR #308, opened 2018-05-09
Static assets directory

andrewosh:
This PR adds support for a custom static assets directory (via the `static` option) as well as a few additional options + cleanup to make it easier to use Idyll programmatically.

Assets stored in the directory can be referenced in the Idyll markup and/or the HTML template file with the `/static` prefix, regardless of the location of the asset direction on disk.

As an example, if Idyll is started with a `{ static: 'my-static-dir/' }` option, and `my-static-dir/` has directory layout:
```
my-static-dir/
  js/
    custom.js
  images/
    test-image.jpg
```
Then the following use-cases are supported:
1.  `index.idl` can contain markup like  `[img src: 'static/images/test-image.jpg /]'`
2.  The HTML template can contain tags that reference static assets:
    `<script type="text/javascript" src="static/js/custom.js"></script>`

Summary of changes:
1. The `static` option is added to support a static asset directory.
2. The `open` option is added to enable/disable automatically opening the page during watch mode.
3. The `stopWatching` method is added to the Idyll object in `idyll-cli` so that Browsersync can be exited cleanly (otherwise the process will never terminate when using watch mode programmatically).
4. In `idyll-cli`, `test/test-project` has been renamed to `test/basic-project` (more descriptive).
5. Since testing for the proper loading of static assets requires loading resources in a headless browser window, the `jsdom` module has been added to `idyll-cli`'s devDependencies.


packages/idyll-cli/package.json
```diff
@@ -39,7 +39,9 @@
     "change-case": "^3.0.1",
     "debug": "^3.1.0",
     "dev-null": "^0.1.1",
+    "ecstatic": "^3.2.0",
```
mathisonian: what's this one?

packages/idyll-cli/src/pipeline/index.js
```diff
@@ -94,6 +98,12 @@ const build = (opts, paths, resolvers) => {
         writeFile(paths.JS_OUTPUT_FILE, output.js),
         writeFile(paths.CSS_OUTPUT_FILE, output.css),
         writeFile(paths.HTML_OUTPUT_FILE, output.html),
+        pathExists(paths.STATIC_DIR).then(exists => {
+          if (exists) {
+            return copy(paths.STATIC_DIR, paths.STATIC_OUTPUT_DIR)
+          }
+          return null;
+        })
```
mathisonian: Just to confirm - this gets triggered during the build, or both build and watch mode?

packages/idyll-cli/test/static-assets/test.js
```diff
@@ -0,0 +1,139 @@
...
+
+test('creates the expected files', () => {
+  expect(projectBuildFilenames).toEqual(EXPECTED_BUILD_FILENAMES);
+})
+
+test('creates the expected HTML', done => {
+  const dom = new JSDOM(projectBuildResults['index.html'], {
+    url: 'http://localhost:3000',
+    referrer: 'http://localhost:3000',
+    runScripts: 'dangerously',
```
mathisonian: :shipit:

andrewosh:
Added an issue to track the removal of `INPUT_DIR` watching here https://github.com/idyll-lang/idyll/pull/308. This should not longer be necessary after this PR is merged, as all static assets should be in the asset dir.

mathisonian:
LGTM ✨

PR #319, opened 2018-05-14
WIP: Moving to subcommands

andrewosh:
Switching over the a subcommand-oriented CLI will make it easier to add options such as `publish` and `create-component` in the near future. This PR extracts the existing build/watch commands in `bin/idyll.js` into a set of subcommand modules in `bin/cmds/`.

Since this is a breaking API change for the CLI tool, it will have to be accompanied by a major version bump.

Other changes:
1. After discussion with @mathisonian, we decided to remove the positional argument for input files, since this is already duplicated by the `-f` argument and will be better supported by a config file soon. 

**Note: This is not yet tested, and is not ready for merging -- just looking for feedback on the command/option structure at this point**


andrewosh:
To give some examples of what the interface looks like aftert this change:

Running `idyll help` produces the following help screen:
```
❯ ./idyll.js help
Usage: idyll <command> [options]

Commands:
  build                              Turn index.idl into output
  create-component [component-name]  Create a new component within your project
  create [project-name]              Create a new project
  publish                            Publish your project to idyll.pub
  watch                              Build then start the development server
                                                                       [default]

Options:
  --help  Show help                                                    [boolean]
```

By default, `idyll` is aliased to the `idyll watch` subcommand, so it produces the familiar build options help:
```
❯ ./idyll.js
idyll.js watch

Options:
  --static                  Directory where static assets are located
                                                             [default: "static"]
  --transform               Custom browserify transforms to apply.
                                                           [array] [default: []]
  --open                    Automatically open the page in the default browser
                            during watch mode.         [boolean] [default: true]
  -m, --components          Directory where components are located
                                                 [array] [default: "components"]
  -c, --css                 Custom CSS file to include in output
  -d, --datasets            Directory where data files are located
                                                               [default: "data"]
  -q, --default-components  Directory where default set of components are
                            located                                      [array]
  -f, --input-file          File containing Idyll source
  -s, --input-string        Idyll source as a string
  -l, --layout              Name of (or path to) the layout to use
                                                               [default: "blog"]
  -n, --no-minify           Skip JS minification                       [boolean]
  -o, --output              Directory where built files should be written
                                                              [default: "build"]
  -r, --no-ssr              Do not pre-render HTML as part of the build[boolean]
  -t, --template            Path to HTML template
  -e, --theme               Name of (or path to) the theme to use
                                                              [default: "idyll"]
  -h, --help                Show help                                  [boolean]
  -p, --port                Custom port to bind the local server to.

Examples:
  idyll.js watch -f index.idl  Build a single .idl file and start the dev server

Not enough non-option arguments: got 0, need at least 1
```

mathisonian:
This is looking super clean. 

For installing / creating a project, we'll expect that the user does something like this, correct?

```sh
$ npm i -g idyll # install latest version of idyll
$ idyll create <my project> # creates a new folder, installs a local copy of idyll
$ idyll watch # triggers the local copy of idyll in watch mode
```

Should the tool give a warning when an update is available? I think there is a package that provides that functionality, e.g. https://www.npmjs.com/package/update-notifier

mathisonian:
While we're making breaking changes, `-f` is a very weird input flag. What about `-i`?

andrewosh:
Indeed, think it's a good plan to switch the file option to `-i`, since `-f` would generally mean force, or something along those lines. 

Notifying on an available update would be excellent too. As for your usage example, since `watch` is the default command, I was imagining the new project workflow would look more like:
```
$ npm i -g idyll
$ idyll create <my-project>
$ cd <my-project>
$ idyll
```

Waddaya think?

mathisonian:
@andrewosh is this ready for review, or should I hold off?

andrewosh:
I'd hold off until I've cleaned up a bit more of the CLI output for building + watching

mathisonian:
This is looking good, but it looks like the tests are failing due to merge output being included in the template package.json

packages/idyll-cli/bin/cmds/create-project.js
```diff
@@ -0,0 +1,175 @@
...
+    let indexPath = p.join(dir, 'index.idyll');
+
+    let packageJson = JSON.parse(await fs.readFile(packagePath));
+    let indexIdyll = await fs.readFile(indexPath, { encoding: 'utf-8' });
+
+    packageJson.name = name;
+    var title = name.split('-').join(' ').replace(/\b\w/g, l => l.toUpperCase())
+
+    await fs.writeFile(packagePath, JSON.stringify(packageJson, null, 2));
+    // TODO: could add more templating.
```
mathisonian: If we do we probably want to add in a lightweight template parser instead of just find/replace

packages/idyll-cli/bin/cmds/create-project.js
```diff
@@ -0,0 +1,175 @@
...
+    await fs.writeFile(indexPath, indexIdyll.replace(/\{\{title\}\}/g, title));
+  }
+
+  async function installDependencies () {
+    return new Promise((resolve, reject) => {
+      let installer = spawn('yarn', ['install'], {
```
mathisonian: Can we check for the existence of yarn first? The current behavior is to fall back to NPM if it isn't available... or maybe we just use npm

packages/idyll-cli/bin/cmds/publish.js
```diff
@@ -0,0 +1,6 @@
+#! /usr/bin/env node
+
+exports.command = 'publish'
+exports.description = 'Publish your project to idyll.pub'
```
mathisonian: 👍

packages/idyll-cli/bin/idyll.js
```diff
@@ -1,90 +1,19 @@
...
+const idyll = getLocalIdyll()
+
+var cmd
+if (!idyll) {
+  cmd = p.join(__dirname, './cli.js');
+} else {
+  cmd = p.join(idyll, '..', '..', 'bin', 'cli.js');
+}
+spawnSync(cmd, process.argv.slice(2), {
+  stdio: 'inherit'
```
mathisonian: This seems like a great way of handling this issue. I'll try it out on a couple projects locally once tests are passing.

packages/idyll-template-projects/README.md
```diff
@@ -0,0 +1,4 @@
+# idyll-layouts
+Layouts that ship by default in new idyll projects.
+
+*See http://idyll-lang.org for complete documentation.*
```
mathisonian: Should be updated to reflect the `idyll-template-projects` purpose. This can probably be borrowed from https://github.com/idyll-lang/generator-idyll

andrewosh: Oops, missed this too. Updated it with some usage info

packages/idyll-cli/bin/cmds/create-project.js
andrewosh: Hmm, it's probably best to stay consistent across users in this case, even though it shouldn't make a difference. I've run into parity issues before -- updated to just use npm

packages/idyll-cli/bin/cmds/create-project.js
andrewosh: Right. Replacing a single field didn't seem to justify it, but adding that dep later + updating the templates wouldn't break anything for existing users, so it'd be a simple future addition. Want to see it in this PR?

mathisonian: Yeah, sounds good to me. No need to add in this PR

mathisonian:
Closing in favor in #333

PR #324, opened 2018-05-26
Added fix for build process hanging

andrewosh:
It's unclear why the build process is currently hanging, but these changes should probably happen anyway:
1.`idyll.build()` should return the build Promise, instead of omitting a return value.
2. The CLI tool should explicitly exit when the build Promise is resolved.

### DIFF
--- a/packages/idyll-cli/src/index.js
+++ b/packages/idyll-cli/src/index.js
@@ -126,7 +126,7 @@ const idyll = (options = {}, cb) => {
       // Leaving the following timing statement in for backwards-compatibility.
       if (opts.debug) console.time('Build Time');
 
-      pipeline.build(opts, paths, resolvers)
+      return pipeline.build(opts, paths, resolvers)
         .then((output) => {
           debug('Build completed');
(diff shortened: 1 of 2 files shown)


andrewosh:
Oops, this is a bit hasty and will probably break watch mode. one sec

andrewosh:
Alright, added a fix that prevents the process from exited if we're in watch mode.

packages/idyll-cli/src/index.js
```diff
@@ -126,7 +126,7 @@ const idyll = (options = {}, cb) => {
       // Leaving the following timing statement in for backwards-compatibility.
       if (opts.debug) console.time('Build Time');
 
-      pipeline.build(opts, paths, resolvers)
+      return pipeline.build(opts, paths, resolvers)
```
mathisonian: I'm not sure we can do this without making an API change, since it short circuits the later `return this`. I'll keep searching for the root cause of the issue here.

andrewosh: Ah you're right, I thought the `this` was part of the top-level `idyll` function return. Bleh

Issue #330, opened 2018-06-12
Table only shows values for the first twenty rows


fhightower:
If you create a table with the following data, you will observe that the table only displays values for the first twenty rows. There are thirty rows as expected, but only the first thirty are filled in.

```
[Table data:`[{'x': 0},
{'x': 1},
{'x': 2},
{'x': 3},
{'x': 4},
{'x': 5},
{'x': 6},
{'x': 7},
{'x': 8},
{'x': 9},
{'x': 10},
{'x': 11},
{'x': 12},
{'x': 13},
{'x': 14},
{'x': 15},
{'x': 16},
{'x': 17},
{'x': 18},
{'x': 19},
{'x': 20},
{'x': 21},
{'x': 22},
{'x': 23},
{'x': 24},
{'x': 25},
{'x': 26},
{'x': 27},
{'x': 28},
{'x': 29}]` /]
```

mathisonian:
Thanks @fhightower - it seems like what was happening is that the table was trying to paginate the data but the "next page" control wasn't being shown, the fixes in #332 should make this behave as you'd expect.

mathisonian:
this is fixed by #332

fhightower:
Wonderful, thanks @mathisonian !

PR #332, opened 2018-06-19
Fix rows getting cut off in table

mathisonian:
Thanks @fhightower for pointing this issue out -

This update fixes logic in the table component so that pagination is shown on tables when there is more data than rows being shown. The issue from #330 was happening because the table component was trying to paginate, but `showPagination` was being set to false in an unintuitive way.

This should make for a much more intuitive experience.


mathisonian:
for reference, here's the updated behavior using the example code from #330 

tab le

PR #333, opened 2018-06-19
Adding `idyll publish` to publish to idyll.pub

andrewosh:
This PR adds support for the `publish` subcommand, which creates a publicly-visible deployment of the current project. It must be run from within a project directory.

All deployments are currently assigned random subdomains of the form `{project-name}-{random-id}`.

Remaining TODOs:
- [x] Configure all the DNS (the API server is currently at `idyll-pub-api.now.sh`, and deployments are not subdomains of `idyll.pub`).
- [ ] Let users specify the build directory to deploy (defaults to `build/`).
- [ ] Ensure that a project is built before attempting the deployment
- [ ] Enforce project size limits

idyll-pub


andrewosh:
Ah, actually there are going to be naming collisions (the now.sh deployment ID isn't sufficient here, because each deployment must still have a unique name). Working on a fix.

mathisonian:
I wonder if it would make sense to have the publish command always trigger a re-build (potentially with a flag to skip), rather than just checking for the existence of the output folder.

mathisonian:
This looks good 👍 the other TODO's can be addressed separately.

PR #365, opened 2018-06-29
Fix windows CLI issue

mathisonian:
This is a fix for a Windows-specific issue with the command line that @Abhilash-Chandran reported. This needs to be tested in a windows environment.


mathisonian:
This is tested in a Windows 10 VM and confirmed working.

PR #376, opened 2018-07-17
Yarn install is the default

domoritz:


mathisonian:
Thanks @domoritz! 

For future reference this behavior is documented at https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Travis-CI-supports-yarn

> If both package.json and yarn.lock are present in the root directory of the repository, we run the following command instead of npm install:

Issue #379, opened 2018-07-20
Incorrect Idyll library is loading


mathisonian:
**Describe the bug**
After creating a project using `idyll create`, running the `idyll` command seems to trigger code loaded from the global `idyll` installation rather than from the local project's `node_modules` folder. I have noticed this specifically when trying to modify code in the data resolver, but suspect it is a systemic issue rather than scoped to this particular file. 

**To Reproduce**
1. install idyll globally `npm i -g idyll`
2. create a new post by running `idyll create`
3. navigate into that post's folder, `cd <post-name>`
4. modify a file in `node_modules/idyll`, for example add a print statement to one of the files.
5. run `idyll`, to new behavior won't be triggered

**Expected behavior**
When running `idyll` in a project folder it should correctly resolve to the local copy of idyll.

**Desktop (please complete the following information):**
 - OS: macOS

/cc @andrewosh

andrewosh:
Hmm, this is definitely an odd one -- just attempted to reproduce it (print statements in `node_modules/idyll`) on Ubuntu and it does work correctly there. Is this using current master or a particular version?

mathisonian:
This is with the latest from npm. Perhaps something weird is happening with me having the global bin linked to a local dev copy of idyll or something, will investigate further.

mathisonian:
I think the problem is that the paths option for `require.resolve()` was only added in node `v8.9.0`, but I'm on a previous version (8.5) so that resolution wasn't happening properly. I'll see if theres a polyfill or something for that, since we support node >= 8.0 otherwise.

mathisonian:
documentation for that is here:  https://nodejs.org/api/modules.html#modules_require_resolve_request_options, under the history tab

mathisonian:
We've hit this elsewhere in the project and already have a dependency on the `resolve` module which is meant to solve the issue, e.g. see https://github.com/idyll-lang/idyll/blob/master/packages/idyll-cli/src/pipeline/parse.js#L83

I'll update the bin resolve logic to use this instead so there's wider support.

mathisonian:
Closed in #380

PR #380, opened 2018-07-20
fix cli resolver for node < 8.9

mathisonian:
This updates the resolution logic for using the local copy of idyll in a project's `node_modules` folder. It updates to use the `resolve` module rather than `require.resolve`, since the `paths` option in resolve is only supported in node 8.9 and greater (https://nodejs.org/api/modules.html#modules_require_resolve_request_options).


mathisonian:
This fixes #379

Issue #338, opened 2018-06-25
Update `Scroller` component to used `position: sticky`


mathisonian:
Update `Scroller` component to used `position: sticky`, for performance wins. This is widely supported now.

mathisonian:
See also https://pudding.cool/process/scrollytelling-sticky/

PR #382, opened 2018-07-21
Sticky scroller

tanalan:
Update for Scroller component to use `position: sticky`

Scroller component does not use `position: sticky`.
Addresses issue #338 

Behavior is pretty similar. There may be subtleties with `position: sticky` that I'm unaware of. For the most part, the behavior appears the same to me.

Nah. I hope not

* **Other information**:
Also updated the idyll-layouts blog's layout css margin to use all 4 explicit margin values.
May be a good idea down the road to also do this for centered as well.


packages/idyll-components/src/scroller.js
```diff
@@ -3,10 +3,10 @@ const { filterChildren, mapChildren } = require('idyll-component-children');
 import TextContainer from './text-container';
 const d3 = require('d3-selection');
 
-
 const styles = {
   SCROLL_GRAPHIC: {
-    position: 'absolute',
+    position: '-webkit-sticky',
+    position: 'sticky',
```
mathisonian: Unfortunately setting multiple styles for a single property doesn't work in React styles like it does in CSS. If you want to set both of these values it has to be done in CSS; although I'm not entirely sure if `-webkit-sticky` is still necessary.

packages/idyll-layouts/src/blog/index.js
```diff
@@ -2,7 +2,10 @@ import createStyles from './styles';
 
 const config = {
   maxWidth: 600,
-  margin: '0 0 0 50px'
+  marginTop: 0,
+  marginRight: 0,
+  marginBottom: 0,
+  marginLeft: '50px'
 };
```
mathisonian: The `centered` layout should also be updated to have the margins set more explicitly like this.

mathisonian:
Thanks @tanalan! This looks great, and performance seems super smooth. I left a couple small comments that should be addressed before this gets merged.

packages/idyll-layouts/src/blog/index.js
tanalan: Updated!

packages/idyll-components/src/scroller.js
tanalan: Hmmm intriguing -- thanks! Was not aware of that :]

Though I will try to incorporate it still, as according to the [w3 definition of position sticky](https://www.w3schools.com/cssref/pr_class_position.asp), the webkit prefix is required for Safari 6.1 and on.

mathisonian: One solution is to toggle a class (e.g. `.idyll-sticky`) in the JavaScript, and then set these styles directly in CSS, this could go in the `idyll-layouts`.

tanalan: Mmk so I tried testing it out in Safari with and without `-webkit-sticky` and Safari's behavior just seems odd, as compared to Chrome. I would describe the behavior as less stable.

But it seems like it's fine without `-webkit-sticky` for now, so I pushed a change to remove it. Though I will look into the toggle class solution you suggested!

packages/idyll-layouts/src/blog/styles.js
```diff
@@ -62,6 +62,11 @@ input {
   width: 100%;
 }
 
+.idyll-scroll-graphic {
+	position: -webkit-sticky;
```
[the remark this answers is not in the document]
tanalan: Gotcha!

mathisonian:
This is looking good! I'll test today, but are there any outstanding issues I should know about before merging @tanalan?

tanalan:
@mathisonian I think I resolved issues I saw yesterday, and I believe it should be good to go! I will double check and test again once I'm on laptop but let me know if you find any issues during testing! :]

mathisonian:
LGTM ✨

PR #387, opened 2018-07-27
Graphic overflow

tanalan:
Bug Fix

This PR seeks to improve that by scaling images to be 100% of the viewport.

For Scroller, inside the Graphic component, images that are larger than the `idyll-scroll-graphic` div would maintain its own height and so didn't scale properly. 

Nah. It just affects images within a Scroller-Graphic component.


tanalan:
Actually hang out on this PR. 
Currently need to fix the `centered` layout because the images under a Graphic component will not be centered xD

tanalan:
Okay, so it appears that in the `centered` layout, everything inside a `Graphic` component is off centered. Will do more testing.

Issue #381, opened 2018-07-20
running `idyll --version` should print the version number


mathisonian:
The command line tool should have a `--version` option (similar to `--help`, which already exists) that simply prints the version number; it would be nice if it is smart enough to print both the local and global version numbers. 

- [x] - Global Version
- [x] - Local Version

mathisonian:
#399 adds the global CLI version. at some point we'll want to modify this to check for the locally installed version as well

mathisonian:
Closed in #481

PR #399, opened 2018-08-04
add --version cli option

BTBTravis:
Attempts to solve issues/381

feature

https://github.com/idyll-lang/idyll/issues/381

This adds the `--version` option to the idyll command as in `idyll --version` 

None

* **Other information**:
  * Right now it simply uses the built in yargs version() implementation so it only returns the cli version number, as of right now `3.6.1` and not the entire apps version. It would be better if it showed both versions. 
  * I thought about adding test but as it uses built in yargs function I think its outside the testable code with the current setup of the tests.


mathisonian:
Thanks @BTBTravis!

PR #410, opened 2018-08-15
Component overlay

tanalan:
# Context
As it stands, while the [docs](https://idyll-lang.org/docs/components) pages has information of the various components, some components don't list out all of the available `props` that a component accepts. This lack of information can limit a author's use of Idyll, not making full use of a component's ability. This may especially be the case with custom components, and outside components. 
It would be pretty nice if authors had a debugging information UI for their page, so that they can easily see components and the `props` of each component.
Lastly, this is likely in the far future, but it would be great for authors to be able to dynamically update their components (and even source code) on the Idyll UI.

# This PR (WIP)
This PR introduces an overlay feature that will be helpful for authors to view debugging information on their components:

idyll-debug-overlay-demo

As it stands, here are some new features/views introduced.
* Borders around each component. I mainly introduced this to debug and see how much of a page a component takes up. This idea/design could be incorporated for a final design.

* A "debug" icon on the top right of each component. The [Idyll-quill](https://files.gitter.im/idyll-lang/Lobby/Zkzj/quill-icon.png) icon is being used, as it makes for a nice "edit" button [:
Upon hovering over the icon (where upon the icon then darkens to show that it's active), information is displayed showing the `props` of the component, as well as the values of each prop.
    * In the future, this debug icon can be pressed and perhaps the source code can appear, be edited, and changes be reflected dynamically on the page.

This PR is *not* ready to be merged at all. Something to be implemented would be to provide a small button or icon that lets an author go into this debugging UI, while otherwise maintaining their original page.


tanalan:
There is also an issue where hovering over a `Range` component does not show the `min` prop, for some odd reason, so looking into that.

tanalan:
Borders will be removed. May just keep them there for debugging purposes for now

mathisonian:
This is looking great @tanalan, and I think is going to be super useful for authors as they are learning Idyll, and it will serve as a building block towards adding editing functionality within the document itself.

Here's a couple quick mocks to show how things could be slightly re-arranged to support more content in the page:

author-assist
author-assist-2

packages/idyll-layouts/src/blog/styles.js
```diff
@@ -71,6 +71,32 @@ input {
...
+  position: relative;
+}
+
+.overlay-button {
+  position: absolute;
+  top: 0;
+  right: 0;
+  opacity: .2;
+  background-color: #E7E3D0;
+  background-image: url('https://files.gitter.im/idyll-lang/Lobby/Zkzj/quill-icon.png');
```
mathisonian: We should update this link to something on our domain. You can add the file to the `static` folder in `idyll-docs` to include it in the next push to `idyll-lang.org`

tanalan: Mmk - done via #414

mathisonian: Now available: https://idyll-lang.org/static/images/quill-icon.png

tanalan:
# Aug 23 Summary of Updates

* No more `React-Tooltip` when hovering over the quill. Rather, we now only show the author view of the components when authors click on the quill button, in which case the view comes up (and closes upon clicking the button again)
  * Additionally, this view is now better formatted than before, where we display relevant information such as the name and docs link of the Component, and then a table showing the props of the component. Though much more work will be done, such as styling the table in a more friendly way, as well as trying to display the current values property (e.g. function code is currently not showing up).
* Upon some discussion with @mathisonian , we decided to drop the overlay naming, and instead adopt an "author view" name for this view (though I suppose a proper name to call this feature is still WIP haha).
* Originally, the implementation details for showing the author view information was in `runtime.js`, but this has now been factored out into an `author-tool.js` component, which is definitely more modular.
* Gif attached below

Update Vid

mathisonian:
!!! looking good @tanalan 

- Can we force the table to take 100% width so we don't get the color wrapping like this?
screen shot 2018-08-23 at 5 02 08 pm

- If the current value is a function or otherwise something that is hard to display in the table, it would be okay to just say _function_. 

I'll play around with it a little more and give any other design feedback soon

megan-vo:
This is just an idea you can ignore if you want -- but is it possible to get the border right when you hover over the Idyll icon? That way the border AND the hidden div don't show up at the same time which might make for a smoother transition.

tanalan:
Yeah good idea, I can do that

tanalan:
Ah, after a good amount of time at this, here's the current progress -- I find that the transition is very smooth, though I've yet to implement the border idea. More work will be done!

sep2transition mov

mathisonian:
@tanalan here are the mocks from our in-person discussion:

*Default state*
overlay-mock1

*Hover state*
overlay-mock2

*Active state*
overlay-mock3

tanalan:
Initial implementation of the aforementioned discussion and design:
* The tooltip disappears clicking and existing the quill icon. This is different from the active state design, mainly because I haven't figured out how to get it to stay upon being in the active state.
  * Will change the size of the tooltip, since it's quite big.
* Using the [description](https://material.io/tools/icons/?style=baseline) (search for 'description') icon for the docs link. Have plans to implement showing "docs" upon hovering over the link.
  * Will likely change the background color or something, as the white color of the docs icon does not fit well with the current scheme

progress

mathisonian:
Updates are looking great @tanalan. 

As you mentioned, I think reducing the padding on the tooltip will help. You might also update the background colors: I think black for the tooltip, and something more subtle than the idyll beige for the details (maybe #efefef).

tanalan:
Did a lot of TODOs today. Current progress has:
* Shadow effect when opening the debug view.
  * Keeping background white because of this -- looks better we believe!
* Table has been decreased in width to 90%. Looks more condensed
* The `type` in the table has a monospace font to appear more data-y and code-y.
* Added a hack to make minimal height for a component (e.g. a `[Display]` component is around like 24px high, so the button (24x24) conflicts with the table upon opening)
  * The hack is essentially trying to make the component at least 40px high upon opening the button.
* docs and icon are next to each other now. In the future, if more icons/links added (like src code), may go with the hover appear effect from before.
  * Though the mouse pointer is not properly vertically aligned with the icon right now when hovering

A bug not sure how to fix at the moment is trying to make the button for a `[Header]` not move when pressing the button.

Also, the React-Tooltip bottom color is still not black, and not sure how to address that atm, but maybe there's a chill solution I'm just currently not seeing atm.

-------------------------------------------------------------------------------

progress

megan-vo:
:O I think the white background and opening transitions are great improvements!

Just an observation: The closing transition is a bit rougher than the opening transition I think. The tables shift up sometimes and the frame transitions aren't as smooth. Not sure if that's something big enough you and Matt want fixed or if it can be worked around.

tanalan:
Ah nice catch! Probly due to having to redraw the page since it pushes
content down. Yikes. Unsure atm how to address

tanalan:
Alright, did some work. @mathisonian Implemented everything we discussed last Tuesday.

Two next steps come to mind:
* The docs icon is still not central, as the icon is 24x24, but the actual doc image takes up less than that (I think 20x20?). So it needs to be cropped.
* Testing this with other components, like `Scroller`, `Fixed`, `Aside`, `Chart`, etc.

packages/idyll-document/src/runtime.js
```diff
@@ -442,6 +441,7 @@ class IdyllRuntime extends React.PureComponent {
 IdyllRuntime.defaultProps = {
   layout: 'blog',
   theme: 'github',
+  authorView: 'false',
```
mathisonian: this should be boolean literal `false` rather than the string `'false'`

packages/idyll-document/src/runtime.js
```diff
@@ -176,8 +176,7 @@ const createWrapper = ({ theme, layout, authorView }) => {
         })
       });
       const metaData = childComponent.type._idyll;
-      // give authorView default false value
-      if (this.isAuthorView && metaData && metaData.props) {
+      if (this.isAuthorView != 'false' && metaData && metaData.props) {
```
mathisonian: always use strict equality checks `===` and `!==` unless there is some good reason for doing otherwise

mathisonian:
We need to get that value to flow from the command line tool to the runtime. Under the hood, the command line tool spits out a JavaScript file that instantiates the runtime here:

https://github.com/idyll-lang/idyll/blob/master/packages/idyll-cli/src/client/build.js#L26-L34

I believe it should be enough to simply add your new option to that dictionary, but you'll want to make sure that the correct value is included in the `__IDYLL_OPTS__` variable, which gets populated here: 

https://github.com/idyll-lang/idyll/blob/master/packages/idyll-cli/src/pipeline/bundle-js.js#L19

tanalan:
I think it's ready!

PR #424, opened 2018-10-02
Give `displayType` key to some components

tanalan:
- [*] The commit message follows our guidelines
- [*] Tests for the changes have been added (for bug fixes / features)
- [*] Docs have been added / updated (for bug fixes / features)

This PR gives the components `Action`, `Display`, `Dynamic`, and `Link` a 'displayType' key to their `_.idyll` metadata to signify that they are inline display elements.

What this does is that it'll help #410 be able to only give an overlay view for block components, and not inline components

Nope


mathisonian:
👍

PR #457, opened 2019-01-31
Abstract out author tool component

tanalan:

Feature update where we can allow a user to provide their own author view component to an IdyllDocument, instead of the current [AuthorTool](https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/src/components/author-tool.js) component.

Also, adds css for box shadow immediately around a component-debug-view when using AuthorTool.

N/A

This is kind WIP, but [IdyllDocument](https://github.com/idyll-lang/idyll/blob/author-view-update/packages/idyll-document/src/index.js) in its `render()` now adds an additional `userViewComponent` prop that lets a user provide their own component  -- if not provided, debugging view defaults to the AuthorTool.

I say kinda WIP because what should happen is that on the aforementioned file, on line 81, it should be `this.props.userViewPlaceholder`, but I'm sorta waiting on logic on [IdyllDesktop](https://github.com/idyll-lang/idyll-desktop) to see exactly how to proceed.

Nada

* **Other information**:
Some of the commits here are shared from the [runtime-features](https://github.com/idyll-lang/idyll/tree/runtime-features) branch, b/c I made this new branch from runtime-features.
But I think that once runtime-features is merged, this will not be an issue.


tanalan:
@mathisonian @megan-vo

packages/idyll-document/src/index.js
```diff
@@ -77,6 +78,7 @@ class IdyllDocument extends React.Component {
           }}
           initialState={this.props.initialState || (this.idyllContext ? this.idyllContext.data() : {})}
           ast={ this.props.ast || this.state.ast }
+          userViewComponent={UserViewPlaceholder}
```
mathisonian: I think we just want to pass down `this.props.userViewComponent` here, otherwise it would always show the placeholder.

packages/idyll-document/src/components/user-view-placeholder.js
```diff
@@ -0,0 +1,18 @@
...
+    return (
+      <div className='placeholder'>
+        {this.props.component}
+        <p>Hello is a placeholder for now !!!!</p>
+      </div>
+    );
+  }
+}
+
+export default UserViewPlaceholder;
```
mathisonian: Can we delete this file?

tanalan:
Gotcha -- addressed your comments @mathisonian -- thank you !

PR #458, opened 2019-01-31
Add fullscreen error message in browser for compiler errors

mathisonian:
Related to #31, this PR adds a fullscreen overlay that displays any errors thrown by the Idyll CLI tool. 

Example below:

screen shot 2019-01-31 at 12 53 46 pm


mathisonian:
The other option is something more like this:

screen shot 2019-01-31 at 1 05 11 pm

mathisonian:
It seems there is a bug in browsersync (https://github.com/BrowserSync/browser-sync/issues/538) that will prevent us from using the second option, so we'll need to stick to the first for now.

mathisonian:
Now I just need to figure out why the tests are failing...

Issue #449, opened 2019-01-03
Markdown parser: Bold and paragraph markdown interaction.


rowanc1:
It looks like there might be a bug with the bold/paragraph interaction in the markdown implementation.

Given:
```
This is **bold text.**

Not a new paragraph.
```

There is no new paragraph after the bold text. However, not ending the paragraph with something bold gives the expected result.
```
This is **bold** text.

New paragraph!
```

Output:
![image](https://user-images.githubusercontent.com/913249/50665750-20f19300-0f6f-11e9-8e04-99cfdfa0603b.png)

Idyll version: `3.12.0`

mathisonian:
Thanks for flagging this @rowanc1, looks like a compiler bug with how the bold parser interacts with newlines.

mathisonian:
This was fixed in https://github.com/idyll-lang/idyll/pull/470 and will go out in the next release. Apologies for the slow fix, and thanks again for reporting @rowanc1.

PR #470, opened 2019-02-28
fix error when bold tag occurs at the end of a paragraph

mathisonian:
This is a bugfix for https://github.com/idyll-lang/idyll/issues/449. The current implementation swallows the expected paragraph break that occurs if a paragraph ends with a bold tag. This fix corrects the issue so that breaks are correctly maintained.

### DIFF
--- a/packages/idyll-compiler/src/lexer.js
+++ b/packages/idyll-compiler/src/lexer.js
@@ -1,39 +1,60 @@
-
 var Lexer = require('lex');
 
-const formatToken = (text) => {
+const formatToken = text => {
   text = text || '';
-  const results = []
+  const results = [];
   results.push('TOKEN_VALUE_START');
-  results.push('"' + text.replace(/\"/g, '&quot;').replace(/\\\[/, '[').replace(/\\\]/, ']') + '"');
+  results.push(
+    '"' +
+      text
+        .replace(/\"/g, '&quot;')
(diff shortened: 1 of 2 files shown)


packages/idyll-compiler/src/lexer.js
```diff
@@ -109,81 +147,132 @@ const lex = function(options) {
...
+    var ret = ['STRONG']
+      .concat(recurse(text, { skipLists: true }))
+      .concat(['STRONG_END']);
     if (trailingSpace) {
-      ret = ret.concat(['WORDS']).concat(formatToken(trailingSpace))
+      if (shouldBreak(trailingSpace)) {
+        ret = ret.concat(['BREAK']);
+      } else {
+        ret = ret.concat(['WORDS']).concat(formatToken(trailingSpace));
+      }
```
mathisonian: The PR for this is a bit messy due to this being the first time running prettier on the lexer code, but the relevant changes are here, and are mirrored in the other rules for `strong` and `em`.

mathisonian:
The PR for this is a bit messy due to this being the first time running prettier on the lexer code, but the relevant changes are here, and are mirrored in the other rules for `strong` and `em`.

https://github.com/idyll-lang/idyll/pull/470/files#diff-a139c5a10afaa4cd29c05bda3898b5f1R186

Issue #309, opened 2018-05-09
Add a new test that performs minification of a project


mathisonian:
To help avoid errors like #124, #155, #307. This would require adding or extending the test suite for `idyll-cli`.

PR #482, opened 2019-03-14
Add a test that uses minification

mathisonian:
Currently there is no test to check that idyll projects minify correctly. This adds one that uses minification to confirm that things compile as expected. It also removes an unused old test.


mathisonian:
Fixes #309

PR #493, opened 2019-05-13
Dynamic: include updateProps in doc API?

ZeningQu:
- [ ] The commit message follows our guidelines

Docs update

It seems like `props.updateProps` is used inside the `Dynamic` component but is not mentioned in the documentation: https://idyll-lang.org/docs/components/default/dynamic

List `updateProps` as one of the `props` for `Dynamic` in the documentation 

I'm not sure if this is intentional though. Maybe you don't want people to pass in callbacks when writing markdown? 

My motivation for adding it was that I want to reuse the Dynamic React component directly (not in markdown)

* **Other information**:

### DIFF
--- a/packages/idyll-components/src/dynamic.js
+++ b/packages/idyll-components/src/dynamic.js
@@ -70,6 +70,12 @@ Dynamic._idyll = {
       example: '100',
       defaultValue: 'none',
       description: 'The maximum value.'
+    },
+    {
+      name: 'updateProps',
+      type: 'function',
+      example: '(x) => console.log(x.value)',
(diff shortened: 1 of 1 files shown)


mathisonian:
Hey @ZeningQu,

The `updateProps` method is a special property that is injected to any component embedded in Idyll  and shouldn't be set directly by users. 

This function can be used by components to send updates back to Idyll's reactive graph, more info on it here https://idyll-lang.org/docs/components/custom

ZeningQu:
Ah I see. It makes sense for Idyll to hide `updateProps` from users. I'll close my PR then. Thank you for the explanation!!

Issue #506, opened 2019-05-27
Decimal values don't work for props in components


kmdupr33:
**To Reproduce**
The docs show that we can use numbers are prop values within `index.idyll` files. However, decimal numbers result in a lexer error. See [this editor for reproduction.](https://idyll-lang.org/editor/f544128e-f463-48de-a219-fca43797cff5)

**Expected behavior**
If non-decimal numbers work, I'd expect decimal numbers to work. (or maybe this is just a documentation issue?)

**Screenshots**
![image](https://user-images.githubusercontent.com/5463526/58431291-19946880-807b-11e9-9af8-41f3872a35f2.png)

**Desktop (please complete the following information):**
 - Browser chrome
 - Version 74.0.3729.169

mathisonian:
Thanks @kmdupr33 - as a workaround `[Range min:0.01 /]` should work, but we should update the matching rule [here](https://github.com/idyll-lang/idyll/blob/master/packages/idyll-compiler/src/lexer.js#L395-L400) to have it work without a leading zero

kmdupr33:
Got it. I can handle that. Pr incoming.

PR #514, opened 2019-05-29
Add support for numbers with leading decmials as prop values (issue #…

kmdupr33:
…506)

Bug fix

#506

Numbers with leading decimals are accepted as valid prop values for components.

No

* **Other information**:
None

### DIFF
--- a/packages/idyll-compiler/src/lexer.js
+++ b/packages/idyll-compiler/src/lexer.js
@@ -393,7 +393,9 @@ const lex = function(options) {
   });
 
   lexer.addRule(/[+\-]?\.?[0-9]+\.?[0-9]*/, function(lexeme) {
-    this.reject = !inComponent;
+    const multiplePeriods =
+      (lexeme.match(new RegExp(/\./, 'g')) || []).length >= 2;
+    this.reject = !inComponent || multiplePeriods;
     if (this.reject) return;
     updatePosition(lexeme);
     return ['NUMBER'].concat(formatToken(lexeme));
--- a/packages/idyll-compiler/src/lexer.js
+++ b/packages/idyll-compiler/src/lexer.js
@@ -392,7 +392,7 @@ const lex = function(options) {
     return ['EXPRESSION'].concat(formatToken(lexeme));
   });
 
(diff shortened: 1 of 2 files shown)


packages/idyll-compiler/src/lexer.js
```diff
@@ -392,7 +392,7 @@ const lex = function(options) {
     return ['EXPRESSION'].concat(formatToken(lexeme));
   });
 
-  lexer.addRule(/[+\-]?[0-9]+\.?[0-9]*/, function(lexeme) {
+  lexer.addRule(/[+\-]?\.?[0-9]+\.?[0-9]*/, function(lexeme) {
     this.reject = !inComponent;
```
mathisonian: This looks like it would match `.123.456` and `.123.`. I think the easiest way to avoid this might be to set `this.reject = true` if there are two decimal points found in the number.

packages/idyll-compiler/test/test.js
```diff
@@ -207,6 +207,22 @@ describe('compiler', function() {
...
+      );
+    });
+    it('should handle numbers with decimals as prop values in components', function() {
+      const input = `[component number:1.1 /]`;
+      const lex = Lexer();
+      const results = lex(input);
+      expect(results.tokens.join(' ')).to.eql(
+        'OPEN_BRACKET COMPONENT_NAME TOKEN_VALUE_START "component" TOKEN_VALUE_END COMPONENT_WORD TOKEN_VALUE_START "number" TOKEN_VALUE_END PARAM_SEPARATOR NUMBER TOKEN_VALUE_START "1.1" TOKEN_VALUE_END FORWARD_SLASH CLOSE_BRACKET EOF'
+      );
+    });
```
mathisonian: 🙏 thanks for adding tests!

kmdupr33:
Oh that’s an embarrassing, careless mistake. I’ll definitely fix that and
maybe add a test to guard against that mistake

mathisonian:
Updates look great - thanks @kmdupr33!

mathisonian:
Per our policy, since you've made a substantive PR, I've sent you an invite to join the Idyll organization. This gives you the ability to submit PRs branched directly off of this repo instead of forking, which can make collaborating easier, and enables some additional privileges on issues, etc.

Issue #490, opened 2019-05-10
Custom formatting for `Dynamic`


suhr:
For example, I want to display note names instead of numbers.

mathisonian:
What would the interaction look like in this case? You would drag, for example, an A note and it becomes B, etc? 

We could introduce a custom format option that maps value onto a label. This would be pretty straightforward to implement, would it work for your use case?

suhr:
> You would drag, for example, an A note and it becomes B, etc?

Yeah.

> We could introduce a custom format option that maps value onto a label. This would be pretty straightforward to implement, would it work for your use case?

If this option is a function, sure.

rowanc1:
I have been working on something similar in my own project:

![hsMIXmd1Bx](https://user-images.githubusercontent.com/913249/57751696-c9dd9680-76a3-11e9-88ae-d0b122770edc.gif)

I think basically looking at the syntax similar to:
```
[Dynamic transform:`(value > 0)? value : 'free'`/]
```

mathisonian:
@codemayank how would you feel about taking this one?

codemayank:
> @codemayank how would you feel about taking this one?

Yes, sure I'll take this one.

mathisonian:
This has been added (see #519) and will be available in the next release. 

See https://idyll.pub/post/dynamic-test-3f0ebd38f8d30970b1da2f89/ for example usage.

rowanc1:
Nice. :)

PR #519, opened 2019-05-30
Add custom formatting for Dynamic Component

codemayank:

Feature Update.

Addresses #490 

Dynamic Component will accept a prop `map` that will map a give value to the `value` prop of the component. e.g.

``[Dynamic value:Var1 min:50 max:100 step:10 map:`['Red', 'Red', 'Blue', 'Green']`/]``
``[Dynamic value:Var2 min:0 max:4 step:1 map:`(Var2 % 2 === 0) ? 'even' : 'odd' ` /]``

This will be the output of above
custom_formatting
 

No.

* **Other information**:

### DIFF
--- a/packages/idyll-components/src/dynamic.js
+++ b/packages/idyll-components/src/dynamic.js
@@ -25,10 +25,25 @@ class Dynamic extends React.PureComponent {
     this.drag(Selection.select(node));
   }
 
-  render() {
-    const { format, value } = this.props;
+  transformValue() {
+    const { format, value, map, min, step } = this.props;
+    const mapOnArray = Array.isArray(map);
+    const mapOnConditions = typeof map === 'string';
+    if (mapOnArray) {
+      const commonDiff = Math.ceil(Math.abs(step));
+      const index = Math.floor((value - min) / commonDiff);
+      return index < map.length ? map[index] : value;
+    }
+    if (mapOnConditions) {
+      return map;
+    }
     const formatter = Format.format(format);
-    return <span className="idyll-dynamic">{formatter(value)}</span>;
+    return formatter(value);
+  }
+
+  render() {
+    const display = this.transformValue();
+    return <span className="idyll-dynamic">{display}</span>;
   }
 }
 
@@ -36,7 +51,8 @@ Dynamic.defaultProps = {
   format: '.2f',
   min: Number.NEGATIVE_INFINITY,
   max: Number.POSITIVE_INFINITY,
-  step: 1
+  step: 1,
+  interval: 0
 };
 
 Dynamic._idyll = {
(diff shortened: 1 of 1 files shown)


packages/idyll-components/src/dynamic.js
```diff
@@ -25,18 +25,34 @@ class Dynamic extends React.PureComponent {
...
-    const { format, value } = this.props;
+  transformValue() {
+    const { format, value, map, min, step } = this.props;
+    const mapOnArray = Array.isArray(map);
+    const mapOnConditions = typeof map === 'string';
+    if (mapOnArray) {
+      const commonDiff = Math.ceil(Math.abs(step));
+      const index = Math.floor((value - min) / commonDiff);
+      return index < map.length ? map[index] : value;
+    }
```
mathisonian: I'm worried this mapping might be a bit too "magical". 

I think it may be more straightforward to accept a function that users can define to specify there own mapping, e.g. 

```
[Dynamic value:x map:`(d) => { return ... }`]
```

packages/idyll-components/src/dynamic.js
```diff
@@ -25,18 +25,34 @@ class Dynamic extends React.PureComponent {
...
+    if (mapOnConditions) {
+      return map;
+    }
     const formatter = Format.format(format);
-    return <span className="idyll-dynamic">{formatter(value)}</span>;
+    return formatter(value);
```
mathisonian: Since both the `format` property and the new `map` property both specify how to transform the value to be displayed, it may make sense to combine these two into one prop.

What about using `format` for both, and we take different actions depending on whether that is a string or a function:

```
// standard string formatting
[Dynamic value:x format:"d" /] 
```


```
// custom formatting function
[Dynamic value:x format:`(d) => { return ... }` /] 
```

packages/idyll-components/src/dynamic.js
codemayank: I have updated this so now the format prop accepts a function.

packages/idyll-components/src/dynamic.js
codemayank: Yes, it's better to pass the format function to the same prop. I have updated this so format prop works as described in your comment.

mathisonian:
After thinking about it a bit more, passing in a reactive expression rather than a function seems more idyllic. 

I pushed an update to make this work, and made a short test post implementing some of @rowanc1's examples. See here:

https://idyll.pub/post/dynamic-test-3f0ebd38f8d30970b1da2f89/

Since its not a function anymore and there isn't an easy way to differentiate it from our existing `format` option, I've changed it back to accept a second prop, `display`.  See the example above to see this in action, I think this seems like a pretty nice API.

rowanc1:
Thanks for implementing this @mathisonian and @codemayank!

I started a similar project before I discovered Idyll, I wrote up a bit of a review of the differences in my thinking/features so far:

https://github.com/ink-components/ink-components/issues/8

I would be very curious on your thoughts on this. Really awesome work on this project/language.

Issue #275, opened 2018-03-18
Control Flow.


mathisonian:
For supporting data-driven reports better we probably need to support mapping over data to output idyll components.

I'd suggest something like this (given input `data = [1, 2, 3 ]`

```

normal text

[each data:d]
   [display var:d /]
[/each]

more normal text

```

should output

```
normal text

1
2
3 

more normal text
```

- [x] - Conditional
- [x] - Switch
- [ ] - Map

mathisonian:
I initially resisted the urge to put control flow types of things into the language definition, and indeed I think the component API is rich enough now that we can generate performant responsive control flow without introducing additional syntax. 

I've used version of these components throughout various projects, and while they are relatively small and simple to implement, the task of doing it yourself each time feels like an unnecessary chore. I'd propose adding the following control flow components:

_conditional_
```
[Conditional if:myVar]
[/Conditional]
```

_switch_
```
[Switch value:myVar ]
   [Case test:"string1"] Content 1 [/Case]
   [Case test:"string2"] Content 2 [/Case]
[/Switch]
```

_map_

```
[Map value:myArrayVar currentValue:iterator ]
  [Item value:iterator /]
[/Map]
```

mathisonian:
`Conditional` was added in https://github.com/idyll-lang/idyll/pull/394

codemayank:
@mathisonian I can take this one and add Switch and Map Components.

mathisonian:
This has been addressed by #524 and #528

PR #524, opened 2019-06-04
Add Switch Component

codemayank:
- [X] Tests for the changes have been added (for bug fixes / features)

Adds new component `[Switch /]`

Addresses #275 

`[Switch /]` component will render a child `[Case /]` component whose test prop is equal to `[Switch /]` components value prop. If none of the `[Case /]` components are matched the default case will be rendered. 
```
[var name:"exp" value:"one" /]
[Switch value:exp]
  [Case test:"one"]Case 1[/Case]
  [Case test:"two"]Case 2[/Case]
  [Case test:"three"]Case 3[/Case]
  [Case test:"four"]Case 4[/Case]
  [Case test:"Default"]Default case[/Case]
[/Switch]

``` 

No

* **Other information**:


packages/idyll-components/src/switch.js
```diff
@@ -0,0 +1,34 @@
+const React = require('react');
+
+class Switch extends React.Component {
+  render() {
+    const { idyll, hasError, updateProps, value, ...props } = this.props;
+
+    if (props.children) {
+      const filterFn = value => ({ props }) =>
+        props.children[0].props.test === value;
+      const matchCase = filterFn(value);
```
mathisonian: Can you do the filter using the `filterChildren` function from https://github.com/idyll-lang/idyll-component-children? This will make it more future-proof in case we change how Idyll wraps components

packages/idyll-components/src/switch.js
```diff
@@ -0,0 +1,34 @@
...
+      const matchDefault = filterFn('Default');
```
mathisonian: This is an edge case, but does this mean we couldn't match for the value `Default`? I wonder if its worth including a `[Default /]` component or `[Switch.Default /]` component to cover this functionality.

packages/idyll-components/src/case.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+Case._idyll = {
+  name: 'Case',
+  tagType: 'open',
+  children: ['Some text'],
+  props: [
+    {
+      name: 'test',
+      type: 'variable',
```
mathisonian: I think this should be `value`

packages/idyll-components/src/switch.js
codemayank: Updated this to use idyll-component-children.

packages/idyll-components/src/case.js
codemayank: updated variable to value.

packages/idyll-components/src/switch.js
codemayank: I think having a default component is better than passing "default" as a string. So I have created a default component.

Also, I am using only one filter function to find both the matched case and default case. And used array destructuring to extract the matched case and default case. This, however, restricts users to always write the default case at the end. I am not clear whether having this restriction is good or bad (as this restriction is not there in javascript) lmk what you think about it.

mathisonian: Having the `default` last seems like a reasonable case but I think there a bug in the logic in the case that a default isn't. I've updated to instead run two filters

PR #528, opened 2019-06-10
Add map component

codemayank:
- [X] Tests for the changes have been added (for bug fixes / features)

Feature

Addresses #275. Adds Map component.

Map component maps over an array to output idyll components. e.g the output of below will be three Display components with values one, two and three.
```
[Map value:`['one', 'two', 'three]` currentValue:'iterator']
[Display value:iterator' /]
[/Map]
```

No

* **Other information**:

### DIFF
--- a/packages/idyll-components/src/map.js
+++ b/packages/idyll-components/src/map.js
@@ -4,14 +4,7 @@ import TextContainer from './text-container';
 
 class Map extends React.Component {
   render() {
-    const {
-      idyll,
-      hasError,
-      updateProps,
-      children,
-      value,
-      currentValue
-    } = this.props;
+    const { children, value, currentValue } = this.props;
 
     if (children) {
       return mapChildren(children, child => {
@@ -24,15 +17,7 @@ class Map extends React.Component {
             }
             return props;
           }, newProps);
-          return (
-            <TextContainer>
-              <div>
-                {React.cloneElement(child, {
-                  ...newProps
-                })}
-              </div>
-            </TextContainer>
-          );
+          return React.cloneElement(child, { ...newProps });
         });
       });
     }
--- a/packages/idyll-components/src/map.js
+++ b/packages/idyll-components/src/map.js
@@ -1,0 +1,63 @@
+const React = require('react');
+const { mapChildren } = require('idyll-component-children');
+import TextContainer from './text-container';
+
+class Map extends React.Component {
+  render() {
+    const {
+      idyll,
+      hasError,
+      updateProps,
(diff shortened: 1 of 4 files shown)


packages/idyll-components/src/map.js
```diff
@@ -0,0 +1,63 @@
...
+          let newProps = Object.assign({}, child.props);
+          newProps = Object.keys(child.props).reduce((props, elm) => {
+            if (props[elm] === currentValue) {
+              props[elm] = val;
+              return props;
+            }
+            return props;
+          }, newProps);
+          return (
+            <TextContainer>
```
mathisonian: I don't think we need this `TextContainer` here

codemayank: I have updated this to remove the TextContainer.

packages/idyll-components/src/map.js
```diff
@@ -0,0 +1,48 @@
...
+
+    if (children) {
+      return mapChildren(children, child => {
+        return value.map(val => {
+          let newProps = Object.assign({}, child.props);
+          newProps = Object.keys(child.props).reduce((props, elm) => {
+            if (props[elm] === currentValue) {
+              props[elm] = val;
+              return props;
+            }
```
mathisonian: Let's remove the references to `currentValue` and just have a hard coded keyword that will be replaced by the mapper, say, `_iterator`.

That way

```
[Map value:`['one', 'two', 'three]`]
  [Display value:'_iterator' /]
[/Map]
```

will work, and I think its a little more straightforward.

mathisonian:
What do you think about renaming to `[Loop]`?

/cc @tanalan @megan-vo

codemayank:
> What do you think about renaming to `[Loop]`?

`[Loop]` would be a good name for this. I'll push the final changes once everyone agrees on the name.

tanalan:
> What do you think about renaming to `[Loop]`?
> 
> /cc @tanalan @megan-vo

`Loop` sounds good to me!

mathisonian:
> [Loop] would be a good name for this. I'll push the final changes once everyone agrees on the name.

Sounds good 👍 let's go with `[Loop]` and I'll merge once the final updates are in

packages/idyll-components/src/map.js
codemayank: Updated this as per your comment.

codemayank:
@mathisonian I have pushed the final changes.

mathisonian:
Thanks @codemayank! ✨

Issue #535, opened 2019-06-21
Header anchors


suhr:
Headers should have anchors and contain links to the anchors.

Also, it would be nice to be also able to specify anchor name manually.

mathisonian:
I agree it would be nice to have these id's inferred automatically and provide an easy linking method (see e.g. how its handled in GitHub README's). 

A quick solution to give an id reference is `[h1 id:"my-title"]My Title[/h1]`.

mathisonian:
To wrap the header elements in links automatically I think the easiest solution would be to add new custom `h[1-6]` components to the standard lib that handle this wrapping.

suhr:
In my project, I implemented [this component](https://github.com/suhr/genkey/blob/master/components/h.tsx) component with [this css](https://github.com/suhr/genkey/blob/master/styles.css#L14).

It works fine, though I'd prefer `[H2][/H2]` instead of `[H n:2][/H]`. A some kind of markdown syntax, like the one discussed [here](https://talk.commonmark.org/t/anchors-in-markdown/247) would be even better.

PR #538, opened 2019-06-25
Generate id attribute for h1 tag

amitbadala:

Added a new component which adds an id attribute to the h1 tag based on the header value.

addresses #535 

Using the component users will be able to attach id attribute to the h1 tag.
User can attach a custom id or it will be auto-generated based on the text of h1 tag

* **Other information**:


mathisonian:
Strange that this is erring out on node 7, I haven't seen that issue before with any of the other tests

package.json
```diff
@@ -7,7 +7,7 @@
   },
   "devDependencies": {
     "husky": "^1.1.3",
-    "lerna": "^2.0.0",
+    "lerna": "^3.15.0",
```
mathisonian: Was this intentional?

yarn.lock
```diff
@@ -2,6 +2,822 @@
 # yarn lockfile v1
 
 
+"@evocateur/libnpmaccess@^3.1.0":
+  version "3.1.0"
```
mathisonian: I'm guessing this is related to the lerna version bump?

package.json
amitbadala: I had setup Lerna separately!

package.json
```diff
@@ -7,7 +7,7 @@
   },
   "devDependencies": {
     "husky": "^1.1.3",
-    "lerna": "^2.0.0",
+    "lerna": "2.0.0",
```
mathisonian: Let's keep the `^`

package.json
```diff
@@ -23,5 +23,6 @@
     "hooks": {
       "pre-commit": "pretty-quick --staged"
     }
-  }
+  },
+  "dependencies": {}
```
mathisonian: No need to add this

packages/idyll-components/src/heading1.js
```diff
@@ -0,0 +1,34 @@
...
+    {
+      name: 'id',
+      type: 'string',
+      example: '"my-id"',
+      description: 'attach id attribute to header'
+    }
+  ]
+};
+
+export default Heading1;
```
mathisonian: Can you update this code so that it can produce headers at an arbitrary level? E.g. Create a function that produces a new header component with a specified level 1-6

packages/idyll-components/src/index.js
```diff
@@ -15,6 +15,7 @@ export { default as Fixed } from './fixed';
 export { default as Float } from './float';
 export { default as Gist } from './gist';
 export { default as Header } from './header';
+export { default as Heading1 } from './heading1';
```
mathisonian: We should export as `H1`, `H2`, ..., `H6`.

yarn.lock
```diff
@@ -2,6 +2,11 @@
 # yarn lockfile v1
```
mathisonian: I don't think `yarn.lock` should need to change for this PR

packages/idyll-components/src/heading1.js
amitbadala: we need to pass extra props with the value of [1-6]
and we can render tag based on that.
what do you think of this approach?

mathisonian: I'd prefer not to require the user to do that so that people who are using the syntax like:

```
# Header 1
## Header 2
... etc
```

will still be able to benefit from these updates

packages/idyll-components/src/index.js
amitbadala: I was thinking of including all the components i:e H1-H6 in a single file, and exporting from that same file or should I create a different file for each component? what do you think?

mathisonian: For compatibility sake I think it would be better if they are in separate files. To avoid repetition you could create a utility function that is used by each of the files to create the exported component. @amitbadala

amitbadala:
idyll-header-anchors

package.json
amitbadala: done!!

package.json
amitbadala: done !!

yarn.lock
amitbadala: done !!

packages/idyll-components/src/h1.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import GenerateHeaders from './generateHeaders';
+
+class H1 extends React.PureComponent {
+  render() {
+    const { children } = this.props;
+    return <GenerateHeaders size="1">{children}</GenerateHeaders>;
```
mathisonian: We should update this so that other properties get passed into the generate headers, e.g. we want to support:

```
[H1 id:"my-custom-id" className:"my-custom-className"] Header Text [/H1]
```

packages/idyll-components/src/generateHeaders.js
```diff
@@ -0,0 +1,18 @@
...
+    .replace(/\s+/g, '-')
+    .toLowerCase();
+};
+
+const GenerateHeaders = props => {
+  const { size, children = [] } = props;
+  const headerText = children.join('');
+  const generatedId = generateId(headerText);
+  const HeaderTag = `h${size}`;
+  return <HeaderTag id={generatedId}>{headerText}</HeaderTag>;
```
mathisonian: This should pass other properties (e.g. `className`) into the returned component, and should only use the generated ID if one isn't provided by the user.

packages/idyll-components/src/h1.js
mathisonian: You can use `<GenerateHeaders size="1" {...props}>{children}</GenerateHeaders>` to pass through all of the props to the generator

amitbadala: Done !!

packages/idyll-components/src/generateHeaders.js
amitbadala: Done!!

mathisonian:
Looks great - awesome work @amitbadala! ✨

Issue #522, opened 2019-05-31
Idyll create should have a "no install" option


mathisonian:
**Is your feature request related to a problem? Please describe.**
Idyll create scaffolds a new idyll post and installs all the dependencies in that folder by default. This is great for most cases, but in a multi-page setup you may already have the dependencies available in the parent directly and don't need to install them per-post

**Describe the solution you'd like**
Add an option to `idyll create` that skips the install step. E.g. `idyll create --no-install`.

PR #539, opened 2019-06-26
feat : Option added to skip installing dependencies 

amitbadala:

Feature update.

Addresses #522

Dependencies were installed by default even if the parent folder had dependencies installed.
User can skip this installation by using flag --no-install. For eg : idyll create --no-install 

No.

* **Other information**:

### DIFF
--- a/packages/idyll-cli/bin/cmds/create-project.js
+++ b/packages/idyll-cli/bin/cmds/create-project.js
@@ -122,16 +122,14 @@ async function createProject(answers) {
       'Copying files from template directory into the target directory',
       copyFiles
     ],
-    ['Configuring post', fillTemplates],
-    ['Installing dependencies', installDependencies]
+    ['Configuring post', fillTemplates]
   ];
-
+  if (isinstallDependencies) {
+    stages.push(['Installing dependencies', installDependencies]);
+  }
   console.log(colors.success(startMessage));
   var spinner;
   try {
-    if (!isinstallDependencies) {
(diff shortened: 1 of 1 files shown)


amitbadala:
Skip dependency installation

image

packages/idyll-cli/bin/cmds/create-project.js
```diff
@@ -123,6 +129,9 @@ async function createProject(answers) {
   console.log(colors.success(startMessage));
   var spinner;
   try {
+    if (!isinstallDependencies) {
+      stages = stages.filter(stage => stage[0] !== 'Installing dependencies');
```
mathisonian: This will break if we change the text of the message to something other than `'Installing dependencies'`. I think a slightly cleaner approach would be to only ever add this stage if the flag is set to `true`, rather than removing it if the flag is `false`.

amitbadala: agreed !

mathisonian:
✨ thanks @amitbadala

mathisonian:
As is the Idyll org policy, since you've successfully submitted a substantial PR, I've invited you to become a member of the organization. 

This will let you submit PRs by branching directly off of the idyll master branch rather than forking, which can make collaborating on pull requests easier, and give you the member tag on comments and your profile page.

amitbadala:
Thank you, Matthew !! :+1:

PR #544, opened 2019-07-02
Remove `width` styles from `.article-header`

joshuahhh:
First of all, there are two different `width` styles here. That can't be right, right?

Second of all: Having `width: 100%;` together with `padding-left: 50px;` means that `.article-header` will be 50px larger than the full width of the browser, so the body will overflow with a horizontal scroll bar. This is bummer. Removing these `width` styles prevents this problem.

(But I can't be sure that there aren't reasons these should be here.)

Thanks!

### DIFF
--- a/packages/idyll-layouts/src/blog/styles.js
+++ b/packages/idyll-layouts/src/blog/styles.js
@@ -25,8 +25,6 @@ body {
 }
 
 .article-header {
-  width: 600px;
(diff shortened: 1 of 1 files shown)


mathisonian:
Yup, this looks much better than the current behavior. Thanks for the PR! ✨ (and sorry for the delay getting back)

Issue #357, opened 2018-06-27
Add a desmos component


mathisonian:
https://www.desmos.com/ is a nice site for "graphing calculator" style functionality. It would be nice to have a wrapper component that let's users easily embed these graphs in idyll articles.

PR #547, opened 2019-07-08
feat : incorporate desmos component

amitbadala:

Add **Desmos** component

Addresses #357 

User can embed _**Desmos**_ graph into idyll document

No

* **Other information**:

### DIFF
--- a/packages/idyll-components/src/desmos.js
+++ b/packages/idyll-components/src/desmos.js
@@ -1,37 +1,17 @@
 import React from 'react';
 
+let desmosGraphCount = 0;
+
 class Desmos extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      id: this.props.id ? this.generateId(this.props.id) : `desmos-graph`
+      id: this.props.id
+        ? this.generateId(this.props.id)
+        : `desmos-${desmosGraphCount++}`
     };
-  }
-
-  generateId(id = '') {
-    return id
-      .toString()
-      .trim()
-      .replace(/\s+/g, '-')
-      .toLowerCase();
-  }
-
-  componentDidMount() {
-    const { apiKey, equation } = this.props;
-    const { id } = this.state;
-    const script = document.createElement('script');
-    script.src = `https://www.desmos.com/api/v1.1/calculator.js?apiKey=${apiKey}`;
-    script.async = true;
-    document.body.appendChild(script);
-    script.onload = () => {
-      var elt = document.getElementById(id);
-      var calculator = window.Desmos.GraphingCalculator(elt);
-      if (equation) {
-        calculator.setExpression({ latex: equation });
-      } else {
-        calculator.setBlank();
-      }
-    };
+    this.elt;
+    this.calculator;
   }
 
   render() {
@@ -49,6 +29,38 @@ class Desmos extends React.Component {
     return (
       <div id={this.state.id} style={{ height, width }} {...attributeProps} />
     );
+  }
+
+  generateId(id = '') {
+    return id
+      .toString()
+      .trim()
+      .replace(/\s+/g, '-')
+      .toLowerCase();
+  }
+
+  componentWillUpdate(nextProps) {
+    const { equation } = nextProps;
+    if (equation !== this.props.equation) {
+      this.calculator.setExpression({ latex: equation });
+    }
+  }
+
+  componentDidMount() {
+    const { apiKey, equation } = this.props;
+    const script = document.createElement('script');
+    script.src = `https://www.desmos.com/api/v1.1/calculator.js?apiKey=${apiKey}`;
+    script.async = true;
+    document.body.appendChild(script);
+    script.onload = () => {
+      this.elt = document.getElementById(this.state.id);
+      this.calculator = window.Desmos.GraphingCalculator(this.elt);
+      if (equation) {
+        this.calculator.setExpression({ latex: equation });
+      } else {
+        this.calculator.setBlank();
+      }
+    };
   }
 }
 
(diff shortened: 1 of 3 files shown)


amitbadala:
Users can now embed **desmos** graph's within their **idyll** projects
for eg :
```[Desmos apiKey:"dcb31709b452b1cf9dc26972add0fda6" equation:"y=x^2" id:"linear graph"/]```
will embed the respective graph in your document .

desmos

**Comment by amitbadala** [comment]
@mathisonian  what do you think ?

**Comment by mathisonian** [comment]
Sick! A couple quick things:

- Can we drop the `id` requirement, and autogenerate a _unique_ id if one isn't provided. 
- Any way to make this responsive if no width is provided? I imagine this might be tricky if Desmos doesn't have built-in support for it, so okay if the answer is no
- What happens if the equation passed in is a reactive expression? e.g. what if I did something like this? Does the Desmos component respond?

```
[var name:"equation" value:"y = x^2" /]

[Desmos apiKey:"..." equation:equation id:"linear graph"/]

[Button onClick:`equation = "y = x^3" `] Change Equation [/Button]
```

amitbadala:
> 1. Can we drop the `id` requirement, and autogenerate a _unique_ id if one isn't provided.
  **Ans : added autogenerated unique id to the component,but getting this error**
```Prop 'id' did not match. Server: "X" Client: "Y"``` **,so wip** 

> 2. Any way to make this responsive if no width is provided? I imagine this might be tricky if Desmos doesn't have built-in support for it, so okay if the answer is no
   **Ans : width is an optional element if the user doesn't provide it will take the width of the parent element, and automatically become responsive.Yes they have a built-in support.
So it is recommended to avoid width although the option should be available** 

> 3. What happens if the equation passed in is a reactive expression? e.g. what if I did something like this? Does the Desmos component respond? 
  **Ans : added this functionality will be committing the same soon.**

mathisonian:
Thanks @amitbadala!

That all sounds good. And I wasn't meaning to rush you with the review comment, just wanted to make sure you saw the earlier note. 

>  Ans : added autogenerated unique id to the component,but getting this error
Prop 'id' did not match. Server: "X" Client: "Y" ,so wip

This error looks related to server side rendering. I'd check to see if maybe the server is using the non-transpiled version of the component to do the initial render, and the client is using an (outdated) transpiled version. If you need to continually build a component while developing, you can run leave `yarn dev running` inside the `idyll-components` package to automatically recompile on changes.

amitbadala:
@mathisonian  I had seen your reviews and started working on it immediately on all the points including 3rd one should've acknowledged it first.
Regarding this error, I am generating unique id in ```componentWillMount``` lifecycle method and after logging the id. It seems it's executing twice, after executing Idyll command you could see one log in Terminal and another in dev tools. As a result demos component unable to locate the id since it references the previous id

mathisonian:
Ah I see. - yes you'll need to figure out a way so that the IDs are generated in a stable order so things stay consistent across server&client.

mathisonian:
I'd recommend creating an ID in the constructor that references a global counter, and increments it each time a new component is instantiated.

amitbadala:
**Equation passed in reactive expressions**

We have two options to load reactive expression in desmos graph.
1. Clear the existing graph and bind a new graph with the new expression.
2. Add the new equation to the existing graph as shown below where the viewer can have control 

Equations
-----------------------------------------------------------------------------------------------------------------------------------------
**Responsiveness** 
 
responsiveDesmos gif
----------------------------------------------------------------------------------------------------------------------------------------
**Autogenerated unique Id** 

autoIdDesmos

amitbadala:
**Questions**

What if the user doesn't have an API key for desmos?
- Can we provide default apiKey?

mathisonian:
Let's go with option (1) for now. It would be ideal if there was a way for us to capture changes if a user edits the equation directly in the component and push that back to the idyll `equation` prop, but I think for the initial version that's not a requirement.

mathisonian:
The other updates look great! 🎉

amitbadala:
Graph re-render on equation change.

desmosEquation

amitbadala:
**Questions** 

1. What if the user doesn't have an API key? we can add a default API key
2. Currently, the desmos API url is embedded in the code. Where can I abstract it?

mathisonian:
> What if the user doesn't have an API key? we can add a default API key

from https://www.desmos.com/api/v1.3/docs/index.html#document-api-keys:
> The demonstration API key, dcb31709b452b1cf9dc26972add0fda6, is provided for use during development. If you plan to use the API in production, you should contact us by e-mailing info@desmos.com to obtain your own API key.

Maybe we should use that one if no API key is provided? We could also print a warning to the console if they are using the default one (if we do show a warning I think it should probably not show if `NODE_ENV` is `production`).

> Currently, the desmos API url is embedded in the code. Where can I abstract it?

Making constant variables at the top of the file should work

amitbadala:
> if we do show a warning I think it should probably not show when building for production

How can we ensure whether it's building for production?

amitbadala:
> if we do show a warning I think it should probably not show when building for production
how can we track that ?

amitbadala:
> if we do show a warning I think it should probably not show when building for production

how can we track that? Need help !

mathisonian:
I guess the easiest way is to just print on the command line and not the JavaScript console. So move the error check from `componentDidMount` to `constructor` (since the mount event never fires on the server). 

Then use `typeof window === 'undefined'` to check if the code is running under node.js rather than in a web browser.

amitbadala:
@mathisonian  done !!

mathisonian:
The updates look great @amitbadala. I just pushed a few minor edits along with code to support communication from the Desmos component back to Idyll. See attached GIF. 

I think its ready to go after these tests pass. Any other updates you wanted to make? 
desmos --  idyll

amitbadala:
>Any other updates you wanted to make?

I think this is good for the basic version, for now, let's see what users think of it and we can work on their feedback accordingly.

PR #570, opened 2019-07-22
Fetch twttr with proper protocol

jprask:
Get protocol from document.location.protocol

I noticed the tweet component was not loading in some cases. Turns out we were using a hardcoded `http` protocol to fetch twttr, causing [mixed content](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content) errors in some cases. This introduces a change to use the correct protocol.

* **What is the current behaviour?** (You can also link to an open issue here)

Always using `http` to fetch twitter script

Use the protocol specified in `document.location.protocol`

No


mathisonian:
Thanks @jprask! LGTM

Issue #509, opened 2019-05-28
Add reference component


mathisonian:
**Is your feature request related to a problem? Please describe.**
A commonly requested component is for references. These are useful for referencing specific external research in a more formal way than just linking to a URL.

**Describe the solution you'd like**
A new component that can be used inline in text for citations and produces a listing of all references at the end of the article. 

This may involve creating two different components, e.g. one for creating citations and one for the listing of references. Example API:

```
Normal text, with something I want to cite [Cite source:"..." /]. 

[References /]
```

Produces:

```
Normal text, with something I want to cite [1]. 

References
[1] - Authors, Title, Etc
```

Note, I'm not sure if it is preferable to use an external resource, e.g. a bibtex file to create a listing of all the references, or to just include them all inline, e.g. `[Cite source:"Authors, Title, Etc" /]` vs `[Cite ref:"ref-name" /]`

amitbadala:
For this, we need to share the state between [Refernces /] and [Cite /]
Was thinking of using Context in React for the same, what do you think @mathisonian?

mathisonian:
Where would you put the context provider in that case?

mathisonian:
I guess you could also put everything in a `cite.js` file and export two separate components:

```
[Cite ref:"ref-name" /]

[Cite.References /]
```

rollaball:
@mathisonian can I take on this issue?

mathisonian:
@rollaball yes! This one will be very useful for people writing technical articles in idyll

rollaball:
We will need to maintain the context state in the parent for this as putting them in the same file still won't make it possible for them to share their states.
Would that be the right approach?

mathisonian:
I still think it would be easiest to keep them in the same file.

Something like:

```js

const citations = {};

class References {
   render() {
     return Object.keys(citations).map(() => {
        return ...
     })
   }
}

class Cite {
  constructor() {
     citations['new-citation'] = {...}
  }

  render() {...}
}

Cite.References = References
module.exports = Cite
```

(this code is just a sketch)

rollaball:
So the problem here would be, as the citation is not being passed as a prop and it is also not inside the state, the references component won't rerender when a new citation is added. So the references would take empty citations array.

mathisonian:
Ah you're right, above would just work if you assume static citations with references at the end of the article. A relatively clean fix for this might be to add a custom setter method on the citations object and force an update on any reference components after a new citation info is added. 

[...]

PR #575, opened 2019-07-31
Add citation and references

rollaball:

Adds Components for Citation and References.

Add [Cite/] anywhere on the page with props authors, title and source.
Then Add [Cite.References /] and it will get the list of all citations on the page

No

* **Other information**:
Issue: #509


rollaball:
```
Idyll is a markup language and toolkit for writing interactive articles. [Cite 
  source:"https://idl.cs.washington.edu/files/2018-Idyll-UIST.pdf" 
  authors:"Matthew Conlen and Jeffrey Heer"
  title:"Idyll: A Markup Language for Authoring and Publishing Interactive Articles on the Web"
/]
![Screenshot from 2019-07-31 15-13-36](https://user-images.githubusercontent.com/14182492/62201851-cd7ce180-b3a5-11e9-9697-e594f90fecbd.png)


[Cite.References/]

```

mathisonian:
This looks great @rollaball! 

One use case this doesn't quite cover yet is a way to cite via an id. For example it is common to cite the same work multiple times, it would be nice if there were a way to provide a shortcut if you want to reference something again, without typing in all the information. Can you think of a clean way to do that?

rollaball:
```[Cite author:"..." title:"..." url:"..." referenceName:"..." /]
[Cite referenceName:"..." /]```
@mathisonian does this seem like a good way to do this?

**Comment by mathisonian** [comment]
That works for me! Only comment is that `referenceName` is a little verbose, what about just `name` or `id`? Of those two `id` maybe is clearer than name since one might reasonable think `name` is referring to the author name or paper title. @fredhohman you have any comment on this?

We should probably print a warning if you provide a `referenceName` but it doesn't have any associated data yet. 

I like this API because it doesn't seem like it would need any changes if we also add bibtex support in the future.

File: packages/idyll-components/src/cite.js [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+    referenceInstances.push(this);
+  }
+  render() {
+    return (
+      <div id="references">
+        <h1>References</h1>
+        <ol>
+          {citations.map((citation, index) => (
+            <li key={index}>
+              {`${citation.authors}: `}
```
mathisonian: Can this be simplified to `{citation.authors}:` or is this a whitespace preservation thing?

fredhohman:
Nice!

I agree `id` is more clear, because `name` could be referring to a number of different fields of a citation/reference/publication.

Other examples:
* I think BibTeX calls it a "citation-key": http://www.bibtex.org/Format/
* Distill calls it a `key` as well. Here's link from their example article template with multiple citations as well: https://github.com/distillpub/template/blob/master/examples/article-test.html#L95

mathisonian:
It should probably accept a `date` property as well that takes an arbitrary string.

mathisonian:
Thanks for the links @fredhohman! 

@rollaball would calling it `key` mess with react's special `key` property?

rollaball:
@mathisonian 
Tried it and it is messing with the special `key`. Will have to call it something else.

rollaball:
@mathisonian 
so what should I go ahead with? `id` or `citation-key`?

mathisonian:
@rollaball let's go with `id`

packages/idyll-components/src/cite.js
rollaball: Replacing it with this to preserve whitespace
```
 {citation.authors}:{' '}
     <a href={citation.url} target="_blank">
```
Or should i remove it all together?

mathisonian: That works 👍

rollaball:
@mathisonian 
I have made the changes

packages/idyll-components/src/cite.js
rollaball: Done

packages/idyll-components/src/cite.js
```diff
@@ -0,0 +1,112 @@
...
+      console.warn(
+        `Warning: Cite with given id not found or invalid id. Check if a Cite with the id "${id}" was declared before this line.`
+      );
+      this.state = {};
+    }
+  }
+  render() {
+    const { citationNumber, authors, title } = this.state;
+    return (
+      <a title={`${authors}: ${title}`} href="#references">
```
mathisonian: Can we hyperlink to the specific citation number?

packages/idyll-components/src/cite.js
```diff
@@ -0,0 +1,112 @@
...
+        <h1>References</h1>
+        <ol>
+          {citations.map((citation, index) => (
+            <li key={index}>
+              {citation.authors}:{' '}
+              <a href={citation.url} target="_blank">
+                {citation.title}
+              </a>
+              , {citation.date}
+            </li>
```
mathisonian: Let's change the format to:

---

[Title with optional link](https://idyll-lang.org), Author Names. Optional Venue. Optional Date. 

---

It should gracefully degrade if properties aren't provided, so e.g. if no date is provided there shouldn't be a double period at the end.

mathisonian: Just made a small edit in the comment above 👆

packages/idyll-components/src/cite.js
rollaball: updated

packages/idyll-components/src/cite.js
rollaball: After change:
Screenshot from 2019-08-04 03-59-16

mathisonian:
Updates look great - thanks @rollaball 🎉

Issue #317, opened 2018-05-14
Asynchronous data loading API


mathisonian:
Syntax like

```
[data name:"..." source:"..." async:true /]
```

would be killer.

manjhawar96:
Questions: 
1. How do components know that this is a data from another source, rather than just treat it like a variable? 
2. What should be the initial data?

rollaball:
So will this data be loaded during component mount or in the build phase?

mathisonian:
The idea is that data would be loaded during the mount. The `[data /]` tag as it is currently implemented brings data in during the build phase, but in certain cases (e.g. with some larger data sets) this becomes a performance issue. So to implement this we'd need to add functionality to the `idyll-runtime` to check for any async requests that need to be made and carry those out. 

We would also need to update the AST schema so that `data` tags with an async prop will validate.

mathisonian:
As @manjhawar96 pointed out - there is the issue of initial data -- I think this should be user specifiable but we should think through what a reasonable default would be (`undefined`? `null`? `[]`?).

rollaball:
If we do pick undefined or null, the components might need to be updated to handle this.
I think user specified or empty array/object might be a good choice.

mathisonian:
That's a good point @rollaball - I think empty array as the default with an optional `initalValue` param would work well

rollaball:
I get the following error in idyll-cli tests when I try to import `csv-parse` in the idyll-document.
 ```
 FAIL  test/minified-project/test.js (31.736s)
  ● Console

    console.error src/index.js:186
      JS_Parse_Error {
        message: 'Unexpected character \'`\'',
        filename: 0,
        line: 8335,
        col: 18,
        pos: 277454 }
```
Any idea why this might be happening?

PR #577, opened 2019-08-02
Async data loading

rollaball:

Data can be fetched asynchronously data when a property `async` is set to `true`

* **Other information**:
Issue: #317

### DIFF
--- a/packages/idyll-document/src/utils/index.js
+++ b/packages/idyll-document/src/utils/index.js
@@ -253,10 +253,19 @@ export const getData = (arr, datasets = {}) => {
       if (typeof fetch !== 'undefined') {
         dataPromise = fetch(sourceValue)
           .then(res => {
+            if (res.status >= 400) {
+              throw new Error(
+                `Error Status ${
+                  res.status
+                } occurred while fetching data from ${sourceValue}. If you are using a file to load the data and not a url, make sure async is not set to true.`
+              );
+            }
             if (sourceValue.endsWith('.csv')) {
               return res
                 .text()
-                .then(resString => parse(resString))
+                .then(resString =>
+                  parse(resString, { cast: true, columns: true })
+                )
                 .catch(e => {
                   console.error(`Error while parsing csv: ${e}`);
                 });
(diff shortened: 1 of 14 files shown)


rollaball:
@mathisonian 
is this good?
added 2 properties to data `async` and `initialValue`.
eg: `[data name:'myAsyncData' source:'url' async:true initialValue:'[]']`

mathisonian:
Sweet that API looks good @rollaball 👍 

I haven't tested the changes locally yet, but the code updates so far look good. Is there any way to do it without adding the `axios` dependency?

rollaball:
@mathisonian , I am not sure how we can do this without package such as axios. What do you recommend?

mathisonian:
Fetch seems like a good solution 👍. This won't work in IE 11, so we should add a note to the docs about that. Users can include their own fetch polyfill if they need async data + IE11 support.

mathisonian:
I'll do a full review of this today.

packages/idyll-document/src/utils/index.js
```diff
@@ -238,12 +239,48 @@ export const filterIdyllProps = (props, filterInjected) => {
...
+
+      let dataPromise = new Promise(res => res(initialValue));
+
+      if (typeof fetch !== 'undefined') {
+        dataPromise = fetch(sourceValue)
+          .then(res => {
+            if (sourceValue.endsWith('.csv')) {
+              return res
+                .text()
+                .then(resString => parse(resString))
```
mathisonian: We should use the same options as the parsing that happens in the CLI so there aren't unexpected differences. There we call `parse(inputString, { cast: true, columns: true })`

packages/idyll-document/src/utils/index.js
```diff
@@ -238,12 +239,48 @@ export const filterIdyllProps = (props, filterInjected) => {
...
+      ? getProperties(val).async.value
+      : false;
+    if (async) {
+      const initialValue = getProperties(val).initialValue
+        ? JSON.parse(getProperties(val).initialValue.value)
+        : [];
+
+      let dataPromise = new Promise(res => res(initialValue));
+
+      if (typeof fetch !== 'undefined') {
```
mathisonian: Can we print a warning if this check fails?

packages/idyll-document/src/utils/index.js
rollaball: updated

packages/idyll-document/src/utils/index.js
rollaball: updated

rollaball:
@mathisonian , I have updated the code and added tests. The tests seem to be failing for some reason as it is unable to fetch the repo in sometimes so could you re-run it once.

mathisonian:
Awesome work @rollaball ! 🎉

PR #586, opened 2019-08-20
Fix switch case and its docs

rollaball:

Fixed Switch component where default case was not working

Default Case doesn't work as we are checking for Boolean of the empty array `matchedCase` as it would always be `true` [here](https://github.com/idyll-lang/idyll/blob/master/packages/idyll-components/src/switch.js#L23)
The live example in the docs had invalid code as well

Fixed the component and the docs


mathisonian:
Thanks @rollaball!

PR #588, opened 2019-08-30
New variable syntax

### LINKED ISSUE #389: variable declaration shorthand
Currently `var`s and `derived`s must be declared one at a time, and each declaration must specify separate `name` and `value` properties. There are also often many `var`s and `derived`s declared near each other at the top of idyll documents. e.g https://github.com/mathisonian/dimensionality-reduction/blob/master/index.idyll#L9-L19.

An alternative would be to allow variables to be defined like:

```
[var 
   varName1:"varValue1" 
   varName2:"varValue2"
  /]
```

we'd have to take care to make sure this is backwards compatible with the existing name/value definitions, but IMO it would make for nice sugar over the current syntax.

rollaball:

allows following syntax to be used to define variables and derived variables.
```
~ x=1, y:=x*2
~ a:=x+y, b="somestring"
```

No

* **Other information**:
Resolves  #389

### DIFF
--- a/packages/idyll-compiler/src/lexer.js
+++ b/packages/idyll-compiler/src/lexer.js
@@ -293,26 +293,27 @@ const lex = function(options, alias = {}) {
     });
     return output.concat(['LIST_END']);
   });
+
   lexer.addRule(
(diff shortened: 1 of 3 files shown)


mathisonian:
This will go in with the next major version bump, so I'm sitting on merging for a bit while we identify other minor and patch level updates that will go out first.

PR #599, opened 2019-11-28
Add byline template

### LINKED ISSUE #593: Delete prefix to author in header component
**Is your feature request related to a problem? Please describe.**
The prefix "By:" to the author property in header component is useful when you write in english but for other language it's a problem.

**Describe the solution you'd like**
Delete the prefix or add a new property to overwrite a default prefix.

jprask:
- [x ] Docs have been added / updated (for bug fixes / features)

First of all sorry this took a while, end of semester got busy :rofl: 

* Add ByLine template to header component. Closes #593 

* By line prefix, suffix & joint is hardcoded

* Header component accepts a `bylineTemplate` prop (optional) that let's the user customize the template.

For example

```
[Header
  title:"Animal Farm"
  subtitle:"Welcome to the jungle"
  authors:`[
    {name: 'Dog', link: 'https://website.com'}, 
    {name: 'Monkey', link: 'https://www.com'},
    {name: 'Duck', link: 'https://www.example.org'}
  ]`
  byLineTemplate: `{
    prefix: 'Made by',
    joint: ' ',
    suffix: '&'
  }`
/]

```

will render the following:

image

No

* **Other information**:

Found a bug where the template is not applied in some context, will be opening a related issue so it can be discussed

### DIFF
--- a/packages/idyll-components/src/header.js
+++ b/packages/idyll-components/src/header.js
@@ -15,13 +15,13 @@ class Header extends React.PureComponent {
         )}
         {this.props.author && (
           <div className={'byline'}>
-            {`${prefix} `}
+            {`${prefix.trim()} `}
             <a href={this.props.authorLink}>{this.props.author}</a>
           </div>
         )}
         {this.props.authors ? (
           <div className={'byline'}>
-            {`${prefix} `}
+            {`${prefix.trim()} `}
             {this.props.authors.map((author, i) => {
               if (typeof author === 'string') {
                 return author;
@@ -31,8 +31,8 @@ class Header extends React.PureComponent {
                   <a href={author.link}>{author.name}</a>
                   {i < this.props.authors.length - 1
                     ? i === this.props.authors.length - 2
-                      ? ` ${suffix} `
-                      : `${joint} `
+                      ? ` ${suffix.trim()} `
+                      : `${joint.trim()} `
                     : ''}
                 </span>
               ) : (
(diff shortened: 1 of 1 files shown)


packages/idyll-components/src/header.js
```diff
@@ -89,6 +94,12 @@ Header._idyll = {
       defaultValue: '"#222"',
       description: 'The background of the header. Can pass a color or a url().'
     },
+    {
+      name: 'byLineTemplate',
+      type: 'object',
+      example: "{ prefix: 'Made by', joint: ' -', suffix: '&' }",
```
mathisonian: Do we need to include any logic around whitespace? Should we always trim user-provided strings, or just assume extra spaces at the start/end of a string is intentional?

mathisonian:
Thanks @jprask, this looks great! 

>  First of all sorry this took a while, end of semester got busy 🤣

Right there with you, apologies for the long review period 😂  I left just one comment, will merge once that looks good (I'm not sure any changes are required, just want to talk through the logic).

packages/idyll-components/src/header.js
jprask: @mathisonian That is a good point, currently it adds a space after the prefix and after the joint as well as a space before and after the suffix, which seems like the most usual spacing. But in the example it would result in `Made by A - B - C & D` which seems okay too. We could either make it more strict by trimming or more customizable by letting all white spaces up to the user. WDYT ?

mathisonian: I think I'm in favor of making it stricter and always rendering "reasonable spaced" output unless you can think of a use case where that would get in the way.

jprask:
@mathisonian Updated this, again sorry it took a while 😄

mathisonian:
🚂 🎉 🎉 

awesome, thank you @jprask !

PR #601, opened 2019-12-25
Added scroll progress option

### LINKED ISSUE #589: Expose scroll progress option
**Is your feature request related to a problem? Please describe.**
Currently the `Scroll` component allows users to trigger updates each time a waypoint is hit, however there is no way to control a variable continuously based on how far into a step the user is. 

**Describe the solution you'd like**
The component should expose a `progress` parameter that utilizes scrollama's `onStepProgress` option.

RakeshUP:

Enhancement for scroller component

There is only step callback in scroller now

Scroller component fires a step progress event now

No

* **Other information**:
Resolves #589

### DIFF
--- a/packages/idyll-components/src/scroller.js
+++ b/packages/idyll-components/src/scroller.js
@@ -53,7 +53,7 @@ class Scroller extends React.Component {
         step: `#idyll-scroll-${this.id} .idyll-step`, // required
         container: `#idyll-scroll-${this.id}`, // required (for sticky)
         graphic: `#idyll-scroll-${this.id} .idyll-scroll-graphic`, // required (for sticky)
-        progress: true
+        progress: this.props.progress != undefined ? true : false
       })
       .onStepEnter(this.handleStepEnter.bind(this))
       .onStepProgress(this.handleStepProgress.bind(this))
@@ -95,8 +95,8 @@ class Scroller extends React.Component {
   }
 
   handleStepProgress(response) {
-    const progress = { response };
-    const update = { progress: progress * 100 };
+    const { progress } = response;
+    const update = { progress };
     this.props.updateProps && this.props.updateProps(update);
   }
 
@@ -212,6 +212,13 @@ Scroller._idyll = {
       example: '`{}`',
       description:
         'The state value associated with the currently selected step. Note you must set the state property on the step components for this value to update.'
+    },
+    {
+      name: 'progress',
+      type: 'integer',
+      example: '0',
+      description:
+        'The percent of completion (0-1) of the currently selected step'
     }
   ]
 };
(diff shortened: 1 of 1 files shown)


RakeshUP:
@mathisonian, I came up with the above solution. Should I change the docs for Scroller component too?

packages/idyll-components/src/scroller.js
```diff
@@ -92,6 +94,12 @@ class Scroller extends React.Component {
     }
   }
 
+  handleStepProgress(response) {
+    const progress = { response };
+    const update = { progress: progress * 100 };
```
mathisonian: At this point the value of `progress` will be `{ response: valueOfTheResponse }`, and so multiplying by 100 will lead to `NaN`.

Currently we have

```js
const progress = { response }
```

which is shorthand for:

```js
const progress = { response: response }
```

but it should instead be
```js
const  { progress } = response

// this is equivalent to:
const progress = response.progress
```

mathisonian: Just a convention thing, but I would not multiply by 100, so the scroll progress is always between `[0, 1]`.

packages/idyll-components/src/scroller.js
```diff
@@ -52,9 +52,11 @@ class Scroller extends React.Component {
       .setup({
         step: `#idyll-scroll-${this.id} .idyll-step`, // required
         container: `#idyll-scroll-${this.id}`, // required (for sticky)
-        graphic: `#idyll-scroll-${this.id} .idyll-scroll-graphic` // required (for sticky)
+        graphic: `#idyll-scroll-${this.id} .idyll-scroll-graphic`, // required (for sticky)
+        progress: true
```
mathisonian: What is the performance implication of this if the user doesn't care about tracking scroll progress? Should we only set `progress` to true if they've bound a variable to the `progress` property?

packages/idyll-components/src/scroller.js
RakeshUP: I didn't check it properly, my bad. Fixed it now!

packages/idyll-components/src/scroller.js
```diff
@@ -52,9 +52,11 @@ class Scroller extends React.Component {
       .setup({
         step: `#idyll-scroll-${this.id} .idyll-step`, // required
         container: `#idyll-scroll-${this.id}`, // required (for sticky)
-        graphic: `#idyll-scroll-${this.id} .idyll-scroll-graphic` // required (for sticky)
+        graphic: `#idyll-scroll-${this.id} .idyll-scroll-graphic`, // required (for sticky)
+        progress: this.props.progress != undefined ? true : false
```
mathisonian: Small thing, but the project style is to always prefer `!==` over `!=` unless there's a specific reason why it isn't appropriate.

packages/idyll-components/src/scroller.js
```diff
@@ -204,6 +212,13 @@ Scroller._idyll = {
       example: '`{}`',
       description:
         'The state value associated with the currently selected step. Note you must set the state property on the step components for this value to update.'
+    },
+    {
+      name: 'progress',
+      type: 'integer',
```
mathisonian: `number` instead of `integer`

packages/idyll-components/src/scroller.js
RakeshUP: Sure. Thanks for pointing out

mathisonian:
🎉

PR #610, opened 2020-01-16
feat: component resolver change

RakeshUP:

Enhancement

Component resolver is initialized at the start of compile process and when a compiler postprocessor plugin creates a new component in components directory, it won't be available till the next build and the current build fails with component not found error.

Component resolver is reloaded after every compile. 
Components directory is removed from browser-sync file watcher when a compiler post-processor is present so that file-watcher doesn't trigger a new build when a component is added to components directory, hence preventing a build loop 

Yes. When users add a compiler plugin, file watcher is disabled for components directory.

* **Other information**:

### DIFF
--- a/packages/idyll-cli/src/index.js
+++ b/packages/idyll-cli/src/index.js
@@ -65,9 +65,7 @@ const idyll = (options = {}, cb) => {
     opts.compiler.postProcessors = opts.compiler.postProcessors.map(
       processor => {
         try {
-          return require(require.resolve(processor, {
-            paths: [paths.INPUT_DIR]
-          }));
+          return require(processor, { basedir: paths.INPUT_DIR });
         } catch (e) {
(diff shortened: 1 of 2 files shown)


mathisonian:
this is great ~ thanks @RakeshUP, this has been an issue for a long time.

mathisonian:
@RakeshUP could you explain a little more why this would be a breaking change vs it being able to get in as a feature version bump

RakeshUP:
@mathisonian, So when a project is set up to use any compiler post-processor, file watcher is disabled for custom components directory in this line https://github.com/idyll-lang/idyll/pull/610/files#diff-f8cdf9c9cdc76d1c6a65da5f2cf424e5R153. 
So it might cause inconvenience. But this won't be a breaking change, just a small change in behavior.

Issue #640, opened 2020-05-14
Add support for multiple configurations in package.json


mathisonian:
**Is your feature request related to a problem? Please describe.**
It is relatively common to want to have two different idyll configurations, one for local development, and one for your production server. 

**Describe the solution you'd like**
There should be a way to specify multiple configurations in the `package.json` file. Something like this could work:

```js
{
  "idyll": [
     ["myDevConfig", { ...configuration options...}],
     ["myProductionConfig", { ...configuration options...}]
  ]
}
```

There would also need to be a way to specify which config you want to use. Something like `idyll --config myDevConfig` or `idyll --env myProductionConfig`. It should default to the first entry if none is specified, and clearly log which config is being used.

mathisonian:
This should be added once https://github.com/idyll-lang/idyll/pull/606 is merged because I believe that will make the implementation a bit easier.

rogerfitz:
Hey I'm starting to work on this. How should it handle multiple configuration files? Say the local config as #606 prioritizes doesn't list dev dependencies but the top level package.json does and dev config was specified via commandline. Is it OK to fail that or should it look for the first valid match when going through priorities? I would think just fail it with an error message like no dev dependencies found in confix file /x/y/z? Could also add an argument to let the user override default error/no error behavior here

mathisonian:
Awesome @rogerfitz! I think failing that with an error message makes sense, its probably better to make things explicit here rather than pass it through and lead to a potentially unexpected configuration. So say we have the following situations:

```
-- enclosing-folder
----  package.json (has dev/prod config)
----  idyll-project
------ package.json (has dev/prod config)
```

^ in this case no error should be thrown, the `package.json` inside of `idyll-project` should take precedence but anything set in the outer `package.json` should be used if it wasn't set by the inner one

```
-- enclosing-folder
----  package.json (has dev/prod config)
----  idyll-project
------ package.json (doesn't have separate dev/prod config)
```

^ in this case an error should be thrown

```
-- enclosing-folder
----  package.json  (doesn't have separate dev/prod config)
----  idyll-project
------ package.json (has dev/prod config)
```

^ in this case no error should be thrown, the `package.json` inside of `idyll-project` should take precedence but property set in the outer package.json is used if isn't set in the inner one



I think this seems like a reasonable pattern to follow, but open to other suggestions. If in doubt prefer making things explicit and throw an error asking the user to fix something.

rogerfitz:
Sounds good! Will go with that

[...]

Issue #694, opened 2021-02-12
Issues with Node v15


mathisonian:
**Describe the bug**
It appears that there may be some incompatibilities with node v15. See https://github.com/idyll-lang/idyll/issues/680 and https://github.com/idyll-lang/idyll/issues/693. 

I'll keep this issue open to track until we've confirmed a fix. For now the recommended solution is to use [https://github.com/nvm-sh/nvm](https://github.com/nvm-sh/nvm) to install another version (e.g. v14) and use idyll with that version.

rogerfitz:
I took a stab at this last night but how do I go about testing `npm install idyll` would work with the branch I'm working on? I got it passing tests with node 15 by upgrading the react-tooltip library. Related issue here https://github.com/npm/cli/issues/1734, `npm install --legacy-peer-deps ..` should work too but the dependency issue was patched in react-tooltip 4.0.1. Creating PR showing what I did here https://github.com/idyll-lang/idyll/pull/706

mathisonian:
If you can run `lerna clean && lerna bootstrap` in the root of the project with node15 then it should be okay on the dependency side. However I noticed that there is also an issue creating new projects: running `idyll create` always says that installing dependencies failed, even when it didn't. 

A little setup required to test the global command line commands with the local dev version of idyll:

- make a new folder
- run `npm init` to create a package json
- inside the new package.json add `scripts:  { "create-idyll-project": "idyll create" }` 
- inside that folder run `yarn link idyll` so that the local copy of idyll is available
- run `yarn run create-idyll-project` to run the idyll create command with the dev version

rogerfitz:
FYI #707 is main PR now

Thanks for these instructions. I was able to get it running but found an issue with yargs and the commandlineArgs library I introduced in idyll-cli. It was saying command "create" not found https://github.com/idyll-lang/idyll/pull/707/commits/386e9cf5d34fed2ce715249d5f0a6fdbf0c8c5a4. Interestingly when I installed idyll 5.3.4 `idyll create` was fine but maybe this is just a yarn link thing?

I added a commit fixing that onto #707 since it was pretty small but would you rather I isolate out the commit to something linked to the issue and another branch/PR? Wasn't sure how detailed you like to be about those sorts of things

mathisonian:
[...]

PR #701, opened 2021-02-22
feat: Add support for multiple configurations in package.json

rogerfitz:

feature

Currently package.json does not support env's and inherits properties following parent vs nested logic introduced in #606

Adds support for package.json idyll to have multiple options. Selectable by API and command line env with functionality as described in #640

No

* **Other information**:
- Did I make the test case too big with the node_modules being included? I can work on trimming that if needed
- Also any thoughts on how best to test the different cases in the test? I did a test where outer-project and inner-project have same environments defined but I wasn't sure how best to have it test the other permutations like we talked about in #640. Either parent has env but nested doesn't and throws error or parent doesn't have env but local does so no error. 
- I modified searchParentDirectories to return the file rather than the file.idyll so I could use things more uniformly between parent file vs local


mathisonian:
Thanks so much for this @rogerfitz! I'm reviewing now and will leave comments as I go. I believe you should be able to safely  exclude the `node_modules` folder from the tests (assuming that all of the dependencies are already available in the project somewhere, if not we can talk through the best way to handle that).

packages/idyll-cli/package.json
```diff
@@ -40,6 +40,7 @@
     "bs-pretty-message": "^1.0.8",
     "chalk": "^2.4.1",
     "change-case": "^3.0.1",
+    "command-line-args": "^5.1.1",
```
mathisonian, line 43: Is this used anywhere?

packages/idyll-docs/pages/docs/configuration-and-styles.js
```diff
@@ -199,6 +199,40 @@ export default ({ url }) => (
       <p>You can also compile an input string directly instead of a file:</p>
       <Highlight className="javascript">{ExampleCodeB}</Highlight>
     </section>
+    <section>
+      <h2>Selecting Idyll options using the env parameter</h2>
```
mathisonian, line 203: 🙏 thanks for updating the docs!

yarn.lock
```diff
@@ -3570,7 +3570,7 @@ command-join@^2.0.0:
   dependencies:
     "@improved/node" "^1.0.0"
 
-command-line-args@^5.0.0:
+command-line-args@^5.0.0, command-line-args@^5.1.1:
```
mathisonian, line 3573: If you end up changing the `package.json` based on my comment above, re-run `yarn` to regenerate this file.

packages/idyll-cli/src/index.js
```diff
@@ -1,330 +1,397 @@
-const fs = require('fs');
```
mathisonian, line 1: Big diff for this file - is this a full re-write or just Git being annoying? If you can point to the parts that changed it will be helpful

mathisonian: Is the change to `searchParentDirectories` the only existing functionality that changed?

packages/idyll-cli/test/env-options/outer-project/node_modules/d3/.gitattributes
```diff
@@ -0,0 +1,2 @@
+d3.js -diff merge=ours
+d3.min.js -diff merge=ours
```
mathisonian, line 2: Let's drop everything in the test's `node_modules`

mathisonian (changes requested):

mathisonian:
There was a small change to `idyll-cli/src/index.js` recently that I think is giving the merge conflict. I believe its just this one liner: https://github.com/idyll-lang/idyll/commit/48b91bfadfe29a8e1bca6d23b2c62d03315e303d#diff-7992088eddfefcd56354965a72d75010b7c7aad4dabaff28a54b378a479ed3ff

packages/idyll-cli/src/index.js
rogerfitz: Yes that and then the new functionality. Will clean up based on your other comment about that one liner

packages/idyll-cli/package.json
rogerfitz: Yes used in index.js to read the env param from command line. Is it OK to use 5.1.1 or should I change it to the one already in yarn.lock 5.0.0? If using 5.1.1 I will re-run yarn.lock like you say to do

mathisonian: Ah okay 👍 keeping `^5.1.1` is fine in that case

packages/idyll-cli/src/index.js
rogerfitz: Just pushed a merged version of this. I think Git was just being funny, I handled the merge by copying idyll master over and making the edits from my fork into that so I didn't miss anything. Should be relatively straightforward to see the line changes but the diff is ugly

yarn.lock
rogerfitz: I think I did this right but not 100% sure. Quite a large diff but that could be expected from full regenerate? I copied yarn.lock from idyll master and then ran yarn install

rogerfitz:
All those comments should be resolved. Let me know if there's anything else that needs changing. 

Anything else you think would be a good issue for me to look at? A while ago I thought I saw something about adding a leaflet component that looked interesting (I'm starting to work with leaflet on some other projects right now) but I could be mistaken. Happy to look deeper into that husky error or windows development issues I had too if that sounds good to you

mathisonian:
Nice, thanks @rogerfitz! I'll review the new changes and get this merged shortly. 

Some type of `[Map /]` component has been at the top of the list, so a leaflet component would be very welcome. There's also a very basic example using [Mapbox](https://mathisonian.github.io/idyll-map-example/).

Another issue that would be helpful is https://github.com/idyll-lang/idyll/issues/694. This one should be a relatively quick fix - it seems like there's an issues with the latest node, I believe related to the specific version of `react-tooltip` that the runtime [depends on](https://github.com/idyll-lang/idyll/blob/master/packages/idyll-document/package.json#L43). The solution likely is either to upgrade that dependency or drop it.

For the husky errors fixes are also welcome. It would also be good to just open an issue on them for visibility if others are running into the same problem

rogerfitz:
Thanks sounds good. I'll take a stab at #694 first and open a husky issue once I replicate it to see if others have the issues too

mathisonian:
Awesome, thanks @rogerfitz! I have a temporary fix for #649 that I'm going to PR and merge soon because I need it for some user studies that are running tomorrow, but if you figure out a way to solve the problem without nuking the dependency completely that would still be appreciated

mathisonian (approved):

mathisonian (approved):

mathisonian:
Just merged, thanks @rogerfitz! Since you've made a substantial contribution I've invited you to join as a member of the Idyll organization on GitHub, per our policy.

In the future you can work off of branches directly on this repo rather than forks, which will make it easier for us to collaborate if we both want to push changes to the same PR.

rogerfitz:
Oh cool thanks!

PR #705, opened 2021-03-05
remove default authorview for node15 compatibility

mathisonian:
This removes the built-in "author view" component that depended on a package version that is incompatible with node 15. cc @rogerfitz

Idyll won't install with `node 15`

Idyll will install with `node 15`

No (it does remove the built in author view component, but this is only for development mode and is still an experimental feature). 

* **Other information**:


mathisonian:
Includes an update to the compiler error handling and tests as well to support node15

Issue #739, opened 2021-07-22
Update react-table dependency in idyll-components to latest


megan-vo:
The Table component in the ```idyll-components``` package is using react-table v6.11, which will no longer have long term support.

We could update react-table to the latest version as well as the Table component to implement the latest API, since it introduces breaking changes.

mrviniciux:
I grab this @mathisonian gonna make a feasibility check first.

mathisonian:
@mrviniciux we just got #740 merged from @megan-vo, so this can be closed now. Now we'll want to upgrade to the latest React across the project so I'll make an issue for that

PR #722, opened 2021-03-31
Bring react table to last v6 version to clear lifecycle warnings. 

NickFoden:
## 🚧 wip 🚧

Currently looking for good table examples to test with, cleaning up the componentWillMount warnings, the react table package is also throwing these warnings - https://github.com/idyll-lang/idyll/issues/711

But wanted to open a draft PR to see if any thoughts on how Idyll is using react-table and also if/when react hooks integration on horizon. 

- I think the api changes for v7 of react-table and will need peer deps of react w/ hooks
#722
<br/><br/>
- **Please check if the PR fulfills these requirements**

* [ ] Tests for the changes have been added (for bug fixes / features)
* [ ] Docs have been added / updated (for bug fixes / features)

- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

* **What is the current behavior?** (You can also link to an open issue here)

- **What is the new behavior (if this is a feature change)?**

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

- **Other information**:

PR #740, opened 2021-07-27
Update react table v7

megan-vo:

* [ ] Tests for the changes have been added (for bug fixes / features)
* [ ] Docs have been added / updated (for bug fixes / features)

Building off of the work from #722, bumps up react-table version to v7 for ```TableComponent```. Behavior should be the same except now the lifecycle warnings should be gone.

A warning for componentWillReceiveProps is logged when using the TableComponent. (Issue #739) 

With the version update, there should be no lifecycle warnings logged for the Table component.

It should not produce a breaking change.

- **Other information**:
Since v7 of react-table is headless and also relies on hooks, we had to switch to using a functional component and construct the JSX elements manually (without using a default Table) as well as do some control of the inputs for page jumping/page sizing options.  I tried breaking up the component into parts to make it more readable and keeping general look of component as close to original, but more than willing to change anything if needed. 

**Test Examples**
Display with full controls
Screen Shot 2021-07-27 at 12 17 53 PM

Display with no pagination
Screen Shot 2021-07-27 at 12 20 29 PM

Display with no rows found
Screen Shot 2021-07-27 at 12 21 00 PM

Sorting
sorting

Next/Previous Buttons
next_prev

Page Jumping
page_jump

Page Size Selection
row_size

Combination of All Above
combo
* Any combination of sorting, jumping to pages, or row size selection will cause the table to jump back to the first page. Can change behavior too if that seems funny.


megan-vo:
Argh looks like some tests are failing because of the underlying HTML structure differences between the two tables. Should I try getting it to match the tests?

mathisonian:
This looks great, thanks @megan-vo! I updated the expected output in the tests so all the checks now pass, and merged this in. It will be included in the next release

PR #771, opened 2022-03-18
Make ids an optional property in the AST

mathisonian:

Update the AST schema so that IDs are optional and not injected by default. Add an option to the compiler and AST converters to specify whether or not IDs should be added

IDs are required in the AST but not really used, making AST changes overly burdensome. 

The AST does not include IDs by default and things work as expected without them.

No

### DIFF
--- a/packages/idyll-ast/src/ast.schema.json
+++ b/packages/idyll-ast/src/ast.schema.json
@@ -65,7 +65,7 @@
                 "type": "string"
               }
             },
-            "required": ["id", "type", "value"],
+            "required": ["type", "value"],
             "additionalProperties": false
           },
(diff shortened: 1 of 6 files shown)


mathisonian:
We'll want to update the AST helpers as well so that they don't require an ID to always be passed.

mathisonian:
cc @jheer @TwoTau

TwoTau:
btw I think line 8 of the `ast.schema.json` should say `description` rather than `desceription`.

mathisonian:
> btw I think line 8 of the ast.schema.json should say description rather than desceription.

fixed, thanks for catching!

PR #775, opened 2022-03-24
Update AST API and transformation pipeline.

jheer:
This PR is a work-in-progress, but ready for preliminary review. I would like to further revise to work with the planned esbuild adoption. So, either that should land and this PR can update or vice versa.

- Update AST API to mutate the input AST nodes. For efficiency, the updated API does not perform schema validation or enforce correctness. Validation can be performed subsequently if desired.
- Update compiler post-processors to use the v2 AST. This change lays the groundwork to work with a future parser that produces a v2 AST directly.
- Unify management of built-in and external custom post-processors.
- Update compiler to always return a Promise, effectively rendering the `async` compiler option useless.
- Fix bug in AST JSON schema specification.
- Update downstream packages (idyll-document, idyll-cli) to use the modified AST API.

This PR should entail major version bumps to idyll-ast and idyll-compiler.


mathisonian:
Thanks @jheer! At first glance these changes look great - lot's of organizational improvements and the updates to the AST API should be very helpful. I'll go through now and do a more thorough review.

packages/idyll-ast/src/ast.schema.json
```diff
@@ -21,7 +21,7 @@
       "description": "The properties/attributes of the element",
       "type": "object",
       "patternProperties": {
-        "^w+": {
+        "^\\w+": {
```
mathisonian, line 24: 👍

packages/idyll-compiler/test/test.js
```diff
@@ -261,17 +270,17 @@ describe('compiler', function() {
   });
 
   describe('parser', function() {
-    it('should parse a simple string', function() {
+    it('should parse a simple string', async function() {
       var input = 'Just a simple string';
-      expect(compile(input, { async: false })).to.eql(
+      expect(await compile(input)).to.eql(
```
mathisonian, line 275: Is the move to the async compiler call just stylistic / convenience? Or is there some functional update here

packages/idyll-ast/test/ast-test.js
```diff
@@ -0,0 +1,269 @@
...
+  //   expect(util.setChildren(createExampleVarNode(), testChildren())).to.eql(
+  //     createExampleVarNode()
+  //   );
+  // });
+
+  // it('should set no children for a data', function() {
+  //   expect(util.setChildren(createExampleData(0), testChildren())).to.eql(
+  //     createExampleData()
+  //   );
+  // });
```
mathisonian, line 126: Should these commented out tests be removed or re-enabled?

packages/idyll-compiler/src/transformers/make-full-width.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    const childName = getNodeName(child).toLowerCase();
+
+    if (
+      [DERIVED, VAR, DATA, FULLWIDTH, SCROLLER].includes(childName) ||
+      hasProperty(child, 'fullWidth')
+    ) {
+      if (childName === FULLWIDTH) {
+        child.name = 'div'; // TODO: use API?
```
mathisonian, line 47: Flagging this `TODO` comment in case there was meant to be an update.

mathisonian (commented):
This looks great! I left a few minor comments. 

Have you tested the full set of changes linked against an article locally? When I try it on my machine I get a `regeneratorRuntime not defined` error coming from the compiler package... though it is that this is due to something in my setup being changed while I was working on `esbuild`. Just checking to see if this is something worth digging more deeply into.

packages/idyll-compiler/test/test.js
jheer: All compile calls return Promises now (all are async), which IMO improves consistency and prevents future errors (e.g., if a caller was using a sync approach but then added an async plugin). So the `await` call (or a `.then`) is necessary for correctness.

packages/idyll-ast/test/ast-test.js
jheer: I can re-enable or swap out for similar. There are also methods that lack tests, and most methods lack comprehensive tests. I plan to improve test coverage as a next step.

packages/idyll-compiler/src/transformers/make-full-width.js
jheer: Good catch. There are no AST API calls for modifications such as changing the component type. The TODO here is a reminder to come back and consider if there should be. After working on this PR further, this seems so rare that I'm thinking we can leave this as-is and remove the comment.

mathisonian: Cool, makes sense to me

mathisonian (approved):
I've updated the compiler babel config to address the issue I mentioned above and removed the unused options from the compiler defaults options object. This LGTM.

PR #776, opened 2022-03-26
Compiler updates

jheer:
**idyll-ast**
- Add extra AST utility methods.
  - `isMetaNode`, `clearProperties`, `getPropertyType`, `getPropertyValue`

**idyll-compiler**
- Replace compiler `options` and `alias` parameters with a single `context` object. The `alias` value is a property of this object, as are any passed in compiler options. This change streamlines the API while making it easy to pass through additional options to the compiler and plugins.
- Add parsed YAML front matter to the `metadata` property of the context object. This allows any parsed front-matter to be available to compiler plugins, and lays groundwork to possibly use front matter as an alternative to the `meta` node.
- Rename `postProcessors` and `transformers` to `plugins`. This results in more consistent and less-verbose naming.

**idyll-cli**
- Fix HTML templates to handle output JS and CSS options. This change fixes an output bug.
- Update compiler to allow function-valued AST plugins. This change allows external libraries to load/resolve their own plugins and pass them in directly without having to go through node module resolution.
- Include the resolved `paths` objects as a property of the `context` object passed to the compiler. This allows plugins to more easily load files (e.g., bibliographic references) based on current project settings.
- Remap the `postProcessors` compiler option to the new `plugins` key for config-file backwards compatibility.


mathisonian (approved):
This looks good to me! The updating handling of compiler options and plugins feels much cleaner, and catching that options bug is very appreciated 🐛 

Seems okay to merge unless there are other changes you want to get in first.

jheer:
No other immediate changes on my end. Thanks!

PR #778, opened 2022-04-04
Allow AST to be passed to build process as option.

jheer:
- Support `ast` option to pass a pre-parsed AST to the Idyll build process.
- Support `inputDir` option to specify an input directory when there is no input file. This option will be ignored if the `inputFile` option is provided. If no `inputFile` is specified, the output `inputString` property will be undefined.

These options allow Idyll to produce bundles for custom input formats, so long as developers can provide an appropriate Idyll AST as input. At this point in time I would not expect these options to be officially supported / documented as part of an Idyll release. Rather they can function as internal "experimental" features to support ongoing research.


packages/idyll-cli/src/pipeline/index.js
```diff
@@ -1,7 +1,7 @@
 const fs = require('fs');
-const Promise = require('bluebird');
```
mathisonian, line 2: Great! It's nice to get rid of these old dependencies now that they're widely available in node.

packages/idyll-cli/src/pipeline/index.js
```diff
@@ -30,115 +30,107 @@ const build = (opts, paths, resolvers) => {
...
-      ]);
-    })
-    .then(() => {
-      return output; // return all results
-    });
+        return null;
+      })
+      .catch(e => {
+        debug('Error copying static files', e);
+      })
```
mathisonian, line 123: I think we may want to just let this error be thrown, so it can be handled at the top level by https://github.com/idyll-lang/idyll/blob/jh/ast-option/packages/idyll-cli/src/index.js#L283-L297

mathisonian: Although I don't have super strong feelings about this if you think handling the errors locally is a better approach.

mathisonian (commented):
Thanks @jheer! 

This looks good to me. I left just one small comment to look at about error handling, this looks good to merge once there's a decision there.

packages/idyll-cli/src/pipeline/index.js
jheer: Up to you! I didn't add this, it just carries over from your prior implementation. Things are just rearranged because I moved the code from a `Promise.then` chain to a "flatter" series of `await` calls.

mathisonian: Ah got it! Then let's keep this PR as it is, that's a separate issue.

Issue #748, opened 2021-10-01
Insert em-dashes when there are three hyphns


mathisonian:
**Is your feature request related to a problem? Please describe.**
It is common for CMSs to automatically replace three hyphens (e.g. `---`) with an em-dash (`—`). 

**Describe the solution you'd like**
Similar to how Idyll will automatically insert smart quotes, it should also convert hyphens to em-dashes, possibly with an option for users to opt-in/out of this behavior. 

This logic should be added to the compiler in the same spot the smartquote logic currently exists (https://github.com/idyll-lang/idyll/blob/master/packages/idyll-compiler/src/processors/post.js#L26-L32).

jimidle:
I was going to have a go at this Matthew, but it seems that post.js is no longer in existence. I will read through the source of the compiler and see if I can make sense of the change logs etc. But if you have time to update the information about this issue, then it make make things a little quicker.

mathisonian:
Thanks @jimidle! Yes - implementing this via a plugin makes sense, I just edited the issue to reflect the recent changes

PR #789, opened 2022-07-11
feat: Implement idyll-lang/idyll#748

jimidle:
- Any text that is not part of a RAW AST node converts 3 --- to em dash
  - Any text that is not part of a RAW AST node converts 2 -- to en dash

* [x] Docs have been added / updated (for bug fixes / features)

Not sure where to put docs for this plugin. It seems that there are no docs for plugins like SmartQuotes? Please advise and I can update this PR to include any. Actually I am not sure how to turn a default plugin on or off?

Feature

#748 

Changes three --- in to an em dash
Changes two -- in to an en dash

Possibly. If default plugins are off by default, then the behavior will not change. However, I think that I have change the index.js such that this plugin is on by default. In which case, this might cause unexpected output changes. Perhaps it should be off by default? But then how do I change the test to turn it on for the test? Add to context?

- **Other information**:

### DIFF
--- a/packages/idyll-compiler/src/index.js
+++ b/packages/idyll-compiler/src/index.js
@@ -15,6 +15,7 @@ import wrapText from './plugins/wrap-text';
 import cleanResults from './plugins/clean-results';
 import smartQuotes from './plugins/smart-quotes';
 import autoLinkify from './plugins/auto-linkify';
+import smartEmEn from './plugins/smart-emen';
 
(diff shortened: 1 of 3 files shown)


mathisonian:
Thanks @jimidle!

I think its okay to turn this on by default even though its a breaking change. It will go out with the 6.0 release which already has some other breaking changes.

